PR #114, opened 2016-06-22
Fix data race on client close

kozlovic:
The queue state object needs to be locked when removing a subscriber
from the list.


mcqueary:
LGTM

PR #118, opened 2016-06-28
Fix durable restart when ack pending >= MaxInFlight

### LINKED ISSUE #117: Durable on restart may not receive messages
If a durable is stopped after the server sent up to `MaxInFlight` messages, when the durable is restarted, it does not receive the pending messages.

kozlovic:
When a durable is restarted, we need to force the delivery, even
if the number of pending acks is greater or equal to the current
MaxInFlight, otherwise, the durable would not receive messages
until after multiple failed redeliveries.

Resolves #117

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1343,7 +1343,8 @@ func (s *StanServer) performDurableRedelivery(sub *subState) {
 		m.Redelivered = true
 
 		sub.Lock()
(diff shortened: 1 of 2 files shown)


ColinSullivan1:
LGTM.  I suggest a code comment regarding revisiting this to look at cleaning up acks up when the durable is closed.

PR #119, opened 2016-07-01
Performance related updates

ColinSullivan1:
- Uses a channel to pass messages to an IO thread
- Uses bufio to flush data to disk.
- fsyncs after a bufio flush to guarantee data


kozlovic:
The branch was initially created with master out-of-date, so had to do a rebase. Since lots of small updates went into this one, I am going to close this PR and resubmit a new one.

PR #121, opened 2016-07-06
Performance Improvements

kozlovic:
Originally submitted by Colin: https://github.com/nats-io/nats-streaming-server/pull/119

- Uses a channel to pass messages to an IO thread
- Uses bufio to flush data to disk (Message and Subscriptions stores)
- fsyncs after a bufio flush to guarantee data


aricart:
Looks good to me

PR #123, opened 2016-07-07
Add CRC checksum for file store

kozlovic:
- Converted Clients file to use protobuf
- Each record has now a CRC-32 checksum field
- Ability to disable CRC check for reads (only occur on startup atm)


mcqueary:
LGTM

PR #126, opened 2016-07-13
[CHANGED] Logging when msgs limit is reached

kozlovic:
We now only print the message once per subject.

### DIFF
--- a/stores/filestore.go
+++ b/stores/filestore.go
@@ -1114,8 +1114,7 @@ func (ms *FileMsgStore) enforceLimits() error {
 		// Remove the first message from our cache
 		if !ms.hitLimit {
 			ms.hitLimit = true
-			Noticef("WARNING: One of the limits (msgs=%v bytes=%v) reached for store [`%s`]",
-				ms.limits.MaxNumMsgs, ms.limits.MaxMsgBytes, ms.subject)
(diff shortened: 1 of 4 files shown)


stores/filestore.go
```diff
@@ -1112,7 +1112,11 @@ func (ms *FileMsgStore) enforceLimits() error {
 		ms.totalBytes -= firstMsgSize
 
 		// Remove the first message from our cache
-		Noticef("WARNING: Removing message[%d] from the store for [`%s`]", ms.first, ms.subject)
+		if !ms.hitLimit {
+			ms.hitLimit = true
+			Noticef("WARNING: One of the limits (msgs=%v bytes=%v) reached for store [`%s`]",
```
ColinSullivan1: WDYT about adding something to the effect that messages will be discarded/unavailable?

kozlovic: You're right. I dropped the fact that old messages are going to be removed from the log statement. Should add something to that effect.

aricart: How about: "Store limits for ['%s'] store have been exceeded. The store will permanently prune old messages as new messages arrive."

ColinSullivan1:
LGTM.

aricart:
LGTM

PR #125, opened 2016-07-13
Add param option to connect to external NATS Server

kozlovic:
By default, the Streaming server embeds the NATS Server.
You can now pass "-embed=false" to connect to an external server.
Still use "-p" or "-a" if needed to connect to the NATS Server.

/cc @ColinSullivan1 @aricart @derekcollison

PR #129, opened 2016-07-14
[CHANGED] Option parameter to connect to external NATS Server

kozlovic:
You can now use `-nats_server <url>` instead. When you provide
the url to connect to a NATS server that requires authentication,
you can pass the username and password in the url itself, or
with `-user` and `-pass`. If username and password is present
in the url, they will override the one passed through the
command line.

This PR is related to #125

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/nats-io/nats"
 	"github.com/nats-io/nats-streaming-server/stores"
 
+	"github.com/nats-io/gnatsd/auth"
 	"io/ioutil"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


aricart:
LGTM

server/server_test.go
```diff
@@ -3295,15 +3303,62 @@ func TestDontEmbedNATSNotRunning(t *testing.T) {
...
+		s := RunServerWithOpts(sOpts, &nOpts)
+		s.Shutdown()
+	}
+
+	notWorkingURLs := []string{
+		"nats://ivan:incorrect@localhost:5223",
+		"nats://localhost:5223,nats://ivan:pwd@localhost:5224",
+		"nats://localhost",
+		"localhost:5223",
+		"localhost",
```
ColinSullivan1: Perhaps a few more to cover bases parsing, `"nats://ivan:pwd@:5224"`, `""`, and `"  "`.

ColinSullivan1:
LGTM

server/server_test.go
kozlovic: @ColinSullivan1 Added 2 out of the 3. The `""` makes the server not use the external server and will therefore try to embed the server, but since we also run the NATS Server manually in this test, the failure to start takes 10 seconds, which is too much.

PR #131, opened 2016-07-17
Fix NATS Streaming Quickstart link [ci skip]

lasselukkari:
I believe that this is the correct link. The current one returns 404 for me.

Any plans or estimates for a Node.js client release? If it's just around the corner I wont bother hacking my own because of the lack of time to do it properly.


kozlovic:
Thanks for the contribution!

PR #139, opened 2016-08-05
[ADDED] Vendoring of all dependencies

### LINKED ISSUE #31: Stan clients may be discarded due to timeout after a reconnect
We changed the HB and Timeout to something like 200ms and 2 seconds, respectively. By default, a NATS client reconnects after 2 seconds. So say that you stop/restart the NATS server (STAN server when NATS is embedded), then when STAN applications reconnect to NATS Server, it may be too late: the STAN server will have purged those clients because of lack of heartbeats.

kozlovic:
Resolves #31


ColinSullivan1:
LGTM.

aricart:
LGTM

PR #141, opened 2016-08-16
[FIXED] Buffered write strategy to reduce risk of partial writes

kozlovic:
Due to buffering, it was possible to flush on disk a partial record.
This increased the risk of failure to recover in case of crash after
a flush operation.
This change ensures that no partial record is stored in the buffer.


ColinSullivan1:
LGTM.

PR #144, opened 2016-08-25
[IMPROVED] Performance by selectively calling Debugf/Tracef

kozlovic:
Debugf/Tracef check an atomic variable to see if the statement
should be logged or not. However, the cost of calling these
functions is significant due to memory allocations caused by
the formatting of the string being logged.
In places where speed matters, this is too much. Using server's
boolean (that capture the configured debug/trace options) and use
those to surround Debugf/Tracef calls improve performance by
limiting the calls to only when debug and/or trace are actually
enabled.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -166,6 +166,16 @@ type StanServer struct {
 	// IO Channel
 	ioChannel     chan (*ioPendingMsg)
 	ioChannelQuit chan bool
+
+	// Use these flags for Debug/Trace in places where speed matters.
+	// Normally, Debugf and Tracef will check an atomic variable to
+	// figure out if the statement should be logged, however, the
+	// cost of calling Debugf/Tracef is still significant since there
+	// may be memory allocations to format the string passed to these
+	// calls. So in those situations, use these flags to surround the
+	// calls to Debugf/Tracef.
+	trace bool
+	debug bool
 }
 
 // subStore holds all known state for all subscriptions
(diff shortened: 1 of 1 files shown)


kozlovic:
Running benchmarks with master and then branch:
```
BenchmarkPublish-8         	   20000	     70054 ns/op	    1419 B/op	      32 allocs/op
BenchmarkPublishAsync-8    	  300000	      4186 ns/op	    1543 B/op	      29 allocs/op
BenchmarkSubscribe-8       	  500000	      3853 ns/op	    1082 B/op	      31 allocs/op
BenchmarkQueueSubscribe-8  	  500000	      4466 ns/op	    1083 B/op	      31 allocs/op
BenchmarkPublishSubscribe-8	  200000	      6469 ns/op	    2619 B/op	      59 allocs/op
```
```
BenchmarkPublish-8         	   20000	     69366 ns/op	    1323 B/op	      28 allocs/op
BenchmarkPublishAsync-8    	  300000	      3929 ns/op	    1437 B/op	      25 allocs/op
BenchmarkSubscribe-8       	 1000000	      2183 ns/op	     637 B/op	      12 allocs/op
BenchmarkQueueSubscribe-8  	 1000000	      3066 ns/op	     635 B/op	      12 allocs/op
BenchmarkPublishSubscribe-8	  300000	      5835 ns/op	    2119 B/op	      38 allocs/op
```
Note the significant improvement, especially on subscribe.

ColinSullivan1:
LGTM.

PR #147, opened 2016-08-29
Fixed some issues with durables

kozlovic:
When a durable was stopped and restarted, everything would be fine.
However, if durable was stopped, then the server restarted (applies
only to server with file store - since with memory based stores,
no state is recovered), when the durable was once again started,
it would get the error: "stan: duplicate durable registration".
This was because the ClientID, which was recovered from the
subscription info, was not cleared from memory, preventing the
durable to restart.


ColinSullivan1:
LGTM

PR #158, opened 2016-09-30
[FIXED] Behavior of queue subscriber when leaving a group

### LINKED ISSUE #157: Issues with Queue Subscribers
- Queue Group is not removed when last queue subscriber is removed.
This means that if there are 5 messages on a channel and you start a queue subscriber on say group "group", this queue subscriber receive all 5 messages (assuming started with `DeliverAllAvailable()`). The queue subscriber `Unsubscribe()` or its connection is closed. The group is not removed from server memory. If you start a new queue subscriber using the same group name, it won't receive any message (again assuming it is created with `DeliverAllAvailable()`).
- A member leaving the group should "transfer" its unacknowledged messages to the remaining queue subscribers in the group. As of now, if you have q1 and q2 on the same group and q1 receives a message and does not ack it, if q1 `Unsubscribe()` or its connection is closed, that message is never redelivered.

kozlovic:
- Transfer all unacknowledged messages to rest of the group.
- When last queue subscriber is removed, group is removed.
- Update group's LastSent on disk if leaving member was
 carrying the value.

Resolves #157

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -304,7 +304,7 @@ func (ss *subStore) updateState(sub *subState) {
 
 // Remove a subscriber from the subscription store, leaving durable
 // subscriptions unless `force` is true.
-func (s *StanServer) Remove(ss *subStore, sub *subState, force bool) {
+func (ss *subStore) Remove(s *StanServer, sub *subState, force bool) {
 	if sub == nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -347,7 +350,45 @@ func (ss *subStore) Remove(sub *subState, force bool) {
...
+			numQSubs := len(qs.subs)
+			idx := 0
+			sub.RLock()
+			for _, m := range sub.acksPending {
+				// Get a remaning queue subscriber.
+				qsub := qs.subs[idx]
+				qsub.Lock()
+				// Give it one of the message.
+				qsub.acksPending[m.Sequence] = m
+				// Make sure we set its ack timer if none already set.
```
kozlovic: Realized that's not enough. The message need to be added to the subscription store (persisted on disk for file stores). Need to revisit this.

server/server.go
```diff
@@ -347,7 +349,74 @@ func (ss *subStore) Remove(sub *subState, force bool) {
...
+		// queue group from the subStore map of queue groups.
+		if empty {
+			delete(ss.qsubs, qgroup)
+		} else {
+			// If there are pending messages in this sub, they need to be
+			// transfered to remaining queue subscribers.
+			// Also, we may need to update the store to keep track of the
+			// group's LastSent if it happens that this leaving member was
+			// carring the LastSent sequence.
+			qs.Lock()
```
kozlovic: You're right. Let me fix that.

kozlovic:
Resolved conflicts and pushed changes from command line. Closing PR.

PR #160, opened 2016-09-30
[FIXED] Remove non durable subscription from store on connection close

### LINKED ISSUE #159: Non durable subscriptions are not removed from storage unless calling Unsubscribe()
Non durable subscriptions are expected to be removed from storage when their connection is closed (that is, no explicit call of `Unsubscribe()`) since the server will not re-create them without the presence of the client/connection info.

kozlovic:
Even without call to Unsubscribe() a non durable subscription should
be removed from the store when its connection is closed.

Resolves #159


ColinSullivan1:
LGTM.

PR #161, opened 2016-10-03
[ADDED] Durable Queue Subscribers

### LINKED ISSUE #156: Durables should work for queue groups
* Ideally, we need a "close" of durable consumer/subscriber that doesn't unsubscribe (right now there is only one unsubscribe). 
* Need design/story for how we store subscription events for queue subscribers (durable or not). Right now we don't have a way to do this
* Add durability to queue subscriber

kozlovic:
- You can now create a durable Queue Subscriber using the DurableName()
  option in the QueueSubscribe() call.
- For durable queue subscribers, QueueSubscribe() will return an error
  if DurableName() contains a name with the ':' character.
- Queue groups with same name can coexist as durable and non durable.
- When creating a subscriber (durable or queue subscriber), the start
  position is ignored for queue groups and existing durable subscribers.
  Delivery will resume from last sent.

Resolves #156

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -98,7 +98,7 @@ var (
 	ErrInvalidUnsubReq = errors.New("stan: invalid unsubscribe request")
 	ErrInvalidCloseReq = errors.New("stan: invalid close request")
 	ErrDupDurable      = errors.New("stan: duplicate durable registration")
-	ErrDurableQueue    = errors.New("stan: queue subscribers can't be durable")
+	ErrInvalidDurName  = errors.New("stan: durable name of a durable queue subscriber can't contain the character ':'")
 	ErrUnknownClient   = errors.New("stan: unkwown clientID")
 )
 
@@ -183,6 +183,7 @@ type subStore struct {
 	qsubs    map[string]*queueState // queue subscribers
 	durables map[string]*subState   // durables lookup
 	acks     map[string]*subState   // ack inbox lookup
+	stan     *StanServer            // back link to Stan server
 }
 
 // Holds all queue subsribers for a subject/group and
@@ -192,6 +193,7 @@ type queueState struct {
 	lastSent uint64
 	subs     []*subState
 	stalled  bool
+	shadow   *subState // For durable case, when last member leaves and group is not closed.
 }
 
 // Holds Subscription state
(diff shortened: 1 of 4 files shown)


server/server_test.go
```diff
@@ -2121,6 +2121,239 @@ func TestStartPositionTimeDelta(t *testing.T) {
...
+		stan.DeliverAllAvailable(),
+		stan.DurableName("dur"),
+		stan.MaxInflight(1)); err != nil {
+		t.Fatalf("Unexpected error on subscribe: %v", err)
+	}
+	// Wait for message to be received and connection to be closed.
+	if err := Wait(ch); err != nil {
+		t.Fatal("Did not get our message")
+	}
+	// Restart a connection
```
ColinSullivan1: From the comment above, did you want to close the connection here?

server/server_test.go
```diff
@@ -2121,6 +2121,239 @@ func TestStartPositionTimeDelta(t *testing.T) {
...
+	if _, err := sc.QueueSubscribe("foo", "group", cb,
+		stan.DeliverAllAvailable(),
+		stan.DurableName("dur"),
+		stan.MaxInflight(1)); err != nil {
+		t.Fatalf("Unexpected error on subscribe: %v", err)
+	}
+	// Wait for message to be received and connection to be closed.
+	if err := Wait(ch); err != nil {
+		t.Fatal("Did not get our message")
+	}
```
ColinSullivan1: Same as above, keep the durable connected, or close the conn?

kozlovic: Yes, if you see in the `cb`, I close the connection when the message with expected sequence is received.

server/server_test.go
kozlovic: Same. In some of the tests, I am closing the connection from the callback. The wait on channel gives me the assurance that both message and connection have been closed.

server/server.go
```diff
@@ -961,8 +998,11 @@ func (s *StanServer) processRecoveredChannels(subscriptions stores.RecoveredSubs
 				if sub.DurableName != "" && !added {
 					sub.ClientID = ""
 				}
-				// Add to the array
-				allSubs = append(allSubs, sub)
+				// Add to the array, unless this is the shadow durable queue sub that
+				// was left in the store in order to maintain the group's state.
+				if !(sub.IsDurable && sub.QGroup != "" && sub.ClientID == "") {
```
kozlovic: Will do.

server/server.go
```diff
@@ -1464,6 +1504,9 @@ func (s *StanServer) performDurableRedelivery(sub *subState) {
 	sortedMsgs := makeSortedMsgs(sub.acksPending)
 	clientID := sub.ClientID
 	durName := sub.DurableName
+	if durName == "" {
```
kozlovic: Yes, good idea.

ColinSullivan1:
LGTM.

PR #163, opened 2016-10-05
Update README.md [ci skip]

njern:
I noticed one of the example commands had a small error.


kozlovic:
@njern Thanks for the contribution. Don't worry about the failed PR. It is due to the testsuite using too much memory and being killed by the system. I am trying to fix that.
In the future, when you make a change to a doc or anything that should not require a build, you can include `[ci skip]` in the summary of your PR. This will skip the build.

njern:
Thanks, that's a good tip.

PR #185, opened 2016-10-27
[ADDED] Connection for subscriber sends

ColinSullivan1:
Publishing quickly to many subscribers could cause the connection from the NATS streaming server to the NATS server to error out with a slow consumer failure.  Based on machine specifications, the number of subscribers causing this error will vary.  

Adding a connection dedicated to sending to subscribers alleviates this, while providing roughly the same performance.  I experimented with additional connections (publish acks, receiving publishes), and surprisingly noticed a performance degradation in our benchmarks.

Because testing is dependent on OS/hardware resources, I did not add a test specific to this case.

On my macbook pro:
` go run stan-bench.go -ns 100 -np 1 -n 10000 -a foo`

would cause the NATS streaming server to error out with:
`[54777] 2016/10/26 20:42:31.180710 [INF] ::1:60417 - cid:2 - Slow Consumer Detected
[54777] 2016/10/26 20:42:31.180725 [INF] ::1:60417 - cid:2 - Slow Consumer Detected`

With these changes the same benchmark now completes to success.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -141,8 +141,13 @@ type StanServer struct {
 	info       spb.ServerInfo // Contains cluster ID and subjects
 	natsServer *server.Server
 	opts       *Options
-	nc         *nats.Conn
-	wg         sync.WaitGroup // Wait on go routines during shutdown
+
+	// For scalability, a dedicated connection is used to publish
+	// messages to subscribers.
+	nc  *nats.Conn // used for most protocol messages
+	ncs *nats.Conn // used solely for publishing to subscribers
+
+	wg sync.WaitGroup // Wait on go routines during shutdown
 
 	// For now, these will be set to the constants DefaultHeartBeatInterval, etc...
 	// but allow to override in tests.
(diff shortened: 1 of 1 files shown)


kozlovic:
LGTM

kozlovic:
If I use the original code I sent you with using the second connection in `ackPublisher` I get better results in pub-only, then sub-only, 1pub-1sub and the above benchmark example. What OS version and Go version are you running on?

ColinSullivan1:
Really...  Interesting.  Yosemite, go 1.7.1

kozlovic:
Yosemite, go 1.6.3 for me. As you know, Go 1.7 performs pretty bad on Yosemite (compare to 1.6), so it would be worth testing on newer OS (asked @mcqueary maybe) or try on different platform.

kozlovic:
Running some tests at the moment on a Linux VM.

ColinSullivan1:
Doing the same on a beefy linux box.

ColinSullivan1:
On linux, I get for this current PR:
```./stan-bench -ns 1 -np 1 -n 5000000 -a foo
Starting benchmark [msgs=5000000, msgsize=128, pubs=1, subs=1]
NATS Streaming Pub/Sub stats: 336,636 msgs/sec ~ 41.09 MB/sec
 Pub stats: 168,328 msgs/sec ~ 20.55 MB/sec
 Sub stats: 168,318 msgs/sec ~ 20.55 MB/sec```

If sending pub acks on the send conn:
``./stan-bench -ns 1 -np 1 -n 5000000 -a foo
Starting benchmark [msgs=5000000, msgsize=128, pubs=1, subs=1]
NATS Streaming Pub/Sub stats: 353,633 msgs/sec ~ 43.17 MB/sec
 Pub stats: 176,828 msgs/sec ~ 21.59 MB/sec
 Sub stats: 176,817 msgs/sec ~ 21.58 MB/sec```

A different result than running locally.  I'll look into this further and update the PR.

**Comment by kozlovic** [comment]
I did not get significant better results using the new connection on both ackPublisher and sendMsgToSub than using it only in sendMsgToSub, but I don't see why performance would degrade using it in both places. The server gets messages from clients, send it to subscribers, *then* send ack back to publishers. There should not be contention using the same connection for subs and publisher acks. However, it should reduce contention between processing incoming messages and sending back acks to publisher, so I would definitively look into using `s.ncs` in `ackPublisher`.

**Comment by ColinSullivan1** [comment]
While locally on my macbook these changes had a negative impact performance, on linux, I found slightly better performance using the send connection to also send publish acknowledgements.  I agree with your thoughts on contention.  I've updated the PR.

**Comment by mcqueary** [comment]
On macOS Sierra 10.12.1
```
stan-bench -ns 1 -np 1 -n 5000000 -a foo
Starting benchmark [msgs=5000000, msgsize=128, pubs=1, subs=1]
NATS Streaming Pub/Sub stats: 293,385 msgs/sec ~ 35.81 MB/sec
 Pub stats: 146,702 msgs/sec ~ 17.91 MB/sec
 Sub stats: 146,699 msgs/sec ~ 17.91 MB/sec
```

PR #190, opened 2016-10-31
update

jmptrader:
update from original


kozlovic:
@jmptrader I don't think that this was meant to be posted on this repo, so closing it now ;-)

Issue #188, opened 2016-10-31
Unsubscribing from Subscription Inside Subscription Handler Errors


sfrooster:
cc @aricart 

Using the node streaming client, we have the following typescript code:
`let subOpts = this.stanConn`
`.subscriptionOptions()`
`.setMaxInFlight(1)`
`.setManualAckMode(true)`
`.setStartWithLastReceived()`
`.setAckWait(ackWaitMS);`
` `
`let lastMsgSub = this.stanConn.subscribe("some-string-here", "", subOpts)`
`    .on("message", (lastMsg: NATSMessage) => {`
`        doReplay(lastMsg);`
`        // we were doing just lastMsgSub.unsubscribe() but go errors`
`        setTimeout(() => lastMsgSub.unsubscribe(), 10000);`
`    });`

The intent is to get just one message, the last received message.  When we had just the commented line (lastMsgSub.unsubscribe()), we routinely got:
`Error: Subject must be supplied`

I tracked it down to https://github.com/nats-io/node-nats-streaming/blob/master/lib/stan.js#L748
where this.subscription.ackInbox was undefined when the error occurs.

I can make the error occur much less frequently with the code above where we defer the unsubscribe for 10 seconds, but it still happens occasionally. I can also reduce the incidence by acking after unsubscribing (but not as much as deferring the unsubscribe). acking before unsubscribing is not preferred for this use case, and also fails nearly as much as just unsubscribing.

aricart:
@sfrooster  What server are you using? There may be an issue with the server forwarding messages before the protocol response is received.

Could you please try with commit hash 624ac895c4065ccaf7cf21496e170945f7e14be6

ColinSullivan1:
@sfrooster We're looking at potential bug introduced in master - I'll keep you posted.

sfrooster:
Ok - so don't try with the commit hash above...

ColinSullivan1:
@sfrooster - Sorry for the confusion - the commit hash @aricart mentioned is good, or give PR #191 a shot.   Which version of the NATS streaming server did you see your problem with?

sfrooster:
Latest - 0.3.1 / 0.9.4

sfrooster:
Given that, which should I be trying?

ColinSullivan1:
If you really want to use master branch, feel free to use commit hash 624ac89.  Otherwise, use release v0.3.0.  Master branch will not be as stable as a release.  Thanks for reporting this!

[...]

PR #191, opened 2016-10-31
Use the send connection to send protocol responses

ColinSullivan1:
Using two connections can theoretically introduce a protocol ordering issue, particularly when subscription responses are published on a different connection than the messages sent to subscribers.

Imagine the following case where a client makes a subscription request.  The NATS streaming server receives this request, and sends a response on the "general" connection, then starts streaming messages on the "send" connection to a subscriber.  The NATS server happens to service the "send" connection first and sends a `MsgProto` to the subscribing client before the `SubscriptionResponse`.  This violates the protocol and results in undefined behavior at the client.

To mitigate this, send any subscription responses on the "send" connection.  It makes sense to be congruent and do the same with publish errors alongside publish acknowledgements.

Performance remains unaffected.  This is nearly impossible to replicate consistently, so there is no test.

This bug may be related to what was seen in #188.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1467,7 +1467,7 @@ func (s *StanServer) processClientPublish(m *nats.Msg) {
 func (s *StanServer) sendPublishErr(subj, guid string, err error) {
 	badMsgAck := &pb.PubAck{Guid: guid, Error: err.Error()}
 	if b, err := badMsgAck.Marshal(); err == nil {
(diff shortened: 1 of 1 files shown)


sfrooster:
This PR seems to have fixed the issue we were having in https://github.com/nats-io/nats-streaming-server/issues/188

sfrooster:
(thumbsup)

PR #192, opened 2016-11-06
Time rotate plus archive

sfrooster:
This is only 80% complete, and  you may not like my approach, but this PR provides the following:

- Specify number of files for a file store
- Specify number of minutes for each file to store messages (each file covers N minutes)
- Allow archiving and specify an archive script to run

The point here is to allow a setup similar to what we'd like:

- Keep 1 week of files online
- Store messages in 1 hour chunks (each file is 1 hour of messages)
- As files become older than 1 week, archive them (move them elsewhere, whatever, you decide) using the specified script which will be called with arguments = topic, temp dat file name, temp idx file name


sfrooster:
example stan.conf:
store_limits: {
    # Override some global limits
    # 1 file per hour
    min_per_file: 60
    # 7 days of files
    num_files: 168
    archive_script: "/Users/mhaan/AppDev/fstore/archive.sh"
}

example archive script:
mkdir -p "../$1"
mv $2 $3 "../$1"

kozlovic:
As we discussed on Slack, we may not go entirely this way, but instead add a way for the user to configure the size of a file slice. The number of slices would no longer be limited to 5. When limits are hit, "empty" slices would be removed (or renamed to `.bak` and script invoked). The file names would have an ever increasing file number, and server will not expect first file to be 1, etc...
Concerns were the potentially high number of files since although only the current slice for writing is kept opened, when caching is disabled, server will have to open/close those slices during messages lookup which will affect performance.

sfrooster:
I like your approach. The one item that's not clear are the options for configuring the size of the file slice - will you be adding a time based option (even if it's not as small as minutes)?

ColinSullivan1:
Thinking aloud, having a process invoke a shell script could violate some users' security policies.  Perhaps having the streaming server move the files to a configurable archive directory might be better.

A script may not even be required if another process is watching the archive directory, although it could be a nice feature.  wdyt?

kozlovic:
@sfrooster Yes, maybe a duration based dimension is possible. Agree with @ColinSullivan1, maybe simply remove or based on configuration rename to `.bak` (and move it?) and leave the responsibility to use to have watchdogs to do what they want with those files.

sfrooster:
That works as long as you're no longer reusing filenames.

sfrooster:
I can appreciate that. One issue was that filenames were being reused but I
understand that would be changed. I think eliminating the script in favor
of moving the files (or just resending them in place) is acceptable - as
long as the user can tell that new files have been archived, what their
names are, which channel they belong to and there won't be any overwriting.

sfrooster:
Resending = renaming

ColinSullivan1:
How about adding a timestamp to the archived filename?

sfrooster:
Well I think now Ivan was talking about continuously increasing the slice
number. Either way, one item to keep in mind is that the logical offshoot
of this is the ability to bring back an archived file and have it
reloaded/made available, so which ever approach supports that I'd say...

sfrooster:
@kozlovic 
Any update on this? :-)

kozlovic:
@sfrooster I don't think that this is something that will be done very quickly. Other priorities at the moment, but you never know.

kozlovic:
Closing in favor of #201

PR #194, opened 2016-11-08
[CHANGED] StartAtSequence and StartAtTimeDelta no longer fail if out …

### LINKED ISSUE #7: Start time/sequence validation should be "inclusive" vs. "exact"
_From @mcqueary on March 15, 2016 19:14_

See [here](https://github.com/nats-io/stan/blob/master/server.go#L1061-L1070) and [here](https://github.com/nats-io/stan/blob/master/server.go#L1226-L1235) in `server.go`.

As a user, if my `SubscriptionRequest` specifies a time 100 seconds in the past, but the earliest stored message was 80 seconds in the past, I get `stan: invalid start time`. IMO that's broken, conceptually. 

Think of how you'd use a "since" date when qualifying Google search results (i.e. show me hits since this date). If I entered 1/1/2016, but the first result was from 1/3/2016, I'd expect to get everything starting from 1/3/2016, and not an error indicating that my "since" date was wrong. The same applies here: if my start time is before the earliest message, just start with the earliest message!

It's tempting to say "well, just use StartPosition.First instead", but I as a user actually require data since exactly (at most) 100 seconds ago, and I have no idea when "First" will start (3 days, 2 hours, or 5 seconds). 

I propose that if start time (or sequence) is earlier than the first message, that we start with the first message, and likewise that if the start time (or sequence) is later than the last message, we set the start time/sequence to 1ns or 1 seqno after the last message (i.e. default to `StartPosition.NewOnly`).

_Copied from original issue: nats-io/stan#24_

kozlovic:
…of range

StartAtSequence will now start delivery at first message if given
sequence is smaller than first available, and equivalent to "NewOnly"
if given sequence is greater than current last sequence.

Similar behavior for StartAtTimeDelta.

Resolves #7

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2600,7 +2600,7 @@ func (s *StanServer) setSubStartSequence(cs *stores.ChannelStore, sub *subState,
 	case pb.StartPosition_SequenceStart:
 		// If there is no message, firstSeq and lastSeq will be equal to 0.
 		firstSeq, lastSeq := cs.Msgs.FirstAndLastSequence()
-		// StarSequence is an uint64, so can't be lower than 0.
+		// StartSequence is an uint64, so can't be lower than 0.
 		if sr.StartSequence < firstSeq {
 			// That translates to sending the first message available.
 			lastSent = firstSeq - 1
@@ -2612,8 +2612,8 @@ func (s *StanServer) setSubStartSequence(cs *stores.ChannelStore, sub *subState,
 			// sequence number.
 			lastSent = sr.StartSequence - 1
 		}
-		Debugf("STAN: [Client:%s] Sending from sequence, subject=%s seq=%d",
-			sub.ClientID, sub.subject, lastSent)
+		Debugf("STAN: [Client:%s] Sending from sequence, subject=%s seq_asked=%d actual_seq=%d",
+			sub.ClientID, sub.subject, sr.StartSequence, lastSent)
 	case pb.StartPosition_First:
 		firstSeq := cs.Msgs.FirstSequence()
 		if firstSeq > 0 {
--- a/server/server.go
+++ b/server/server.go
@@ -2365,28 +2365,6 @@ func (s *StanServer) processSubscriptionRequest(m *nats.Msg) {
 		// subscriber re-joining a group that was left with pending messages.
 		err = s.updateDurable(ss, sub)
 	} else {
-		if setStartPos {
-			// Check SequenceStart out of range
-			if sr.StartPosition == pb.StartPosition_SequenceStart {
-				if !s.startSequenceValid(cs, sr.Subject, sr.StartSequence) {
-					Debugf("STAN: [Client:%s] Invalid start sequence in subscription request from %s.",
-						sr.ClientID, m.Subject)
-					s.sendSubscriptionResponseErr(m.Reply, ErrInvalidSequence)
-					return
-				}
-			}
-			// Check for SequenceTime out of range
-			if sr.StartPosition == pb.StartPosition_TimeDeltaStart {
-				startTime := time.Now().UnixNano() - sr.StartTimeDelta
-				if !s.startTimeValid(cs, sr.Subject, startTime) {
-					Debugf("STAN: [Client:%s] Invalid start time in subscription request from %s.",
-						sr.ClientID, m.Subject)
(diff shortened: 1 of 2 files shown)


sfrooster:
@kozlovic 
When is this scheduled to be merged? :-)

kozlovic:
@aricart @ColinSullivan1 @mcqueary If you would like to have a look before merging?

server/server.go
```diff
@@ -2632,14 +2587,29 @@ func (s *StanServer) setSubStartSequence(cs *stores.ChannelStore, sub *subState,
...
+			// so this would translate to "new only" semantic.
 			lastSent = seq - 1
 		}
 		Debugf("STAN: [Client:%s] Sending from time, subject=%s time=%d seq=%d",
 			sub.ClientID, sub.subject, startTime, lastSent)
 	case pb.StartPosition_SequenceStart:
-		if sr.StartSequence > 0 {
+		// If there is no message, firstSeq and lastSeq will be equal to 0.
+		firstSeq, lastSeq := cs.Msgs.FirstAndLastSequence()
+		// StarSequence is an uint64, so can't be lower than 0.
```
ColinSullivan1: Typo in comment.

server/server.go
```diff
@@ -2632,14 +2587,29 @@ func (s *StanServer) setSubStartSequence(cs *stores.ChannelStore, sub *subState,
...
+		if sr.StartSequence < firstSeq {
```
ColinSullivan1: I'd suggest a debug message.  Perhaps even an option to dictate behavior here?  The concern lies in the case where a log rollover occurred - some users would want to know they are not receiving all messages in their stream.

server/server_test.go
```diff
@@ -2090,42 +2090,81 @@ func TestStartPositionFirstSequence(t *testing.T) {
...
-	// As of now, since there is no message, the call will fail.
-	sub, err := sc.Subscribe("foo", cb, stan.StartAtSequence(0))
-	if err == nil {
-		sub.Unsubscribe()
-		t.Fatal("Expected error on subscribe, got none")
+	rch := make(chan bool)
+	// Create subscriber with sequence below firstSeq, it should not fail
+	// and receive message with sequence == firstSeq
+	sub, err := sc.Subscribe("foo", func(m *stan.Msg) {
+		if m.Sequence == firstSeq {
```
ColinSullivan1: I believe you should also check that the first message received is the firstSeq.  Grated, other tests ensure the store has rolled of messages, but it'd be nice.

server/server_test.go
```diff
@@ -2090,42 +2090,81 @@ func TestStartPositionFirstSequence(t *testing.T) {
...
+	if err := Wait(rch); err != nil {
+		t.Fatal("Did not get our message")
+	}
+	sub.Unsubscribe()
 
-	// Send a message now.
+	// Create subscriber with sequence higher than lastSeq, it should not
+	// fail, but not receive until we send a new message.
+	sub, err = sc.Subscribe("foo", func(m *stan.Msg) {
+		if m.Sequence == lastSeq+1 {
```
ColinSullivan1: You'll want to also check to see it's the first message received by the subscriber.

server/server_test.go
```diff
@@ -2137,39 +2176,68 @@ func TestStartPositionTimeDelta(t *testing.T) {
...
 	if err := sc.Publish("foo", []byte("msg2")); err != nil {
 		t.Fatalf("Unexpected error on publish: %v", err)
 	}
 
-	// Start a subscriber with "TimeDelta" as start position.
-	sub, err := sc.Subscribe("foo", cb, stan.StartAtTimeDelta(1*time.Second))
+	// Create subscriber with TimeDelta in the past, should
+	// not fail and get first message.
+	sub, err := sc.Subscribe("foo", func(m *stan.Msg) {
+		if m.Sequence == 1 {
```
ColinSullivan1: ...check that it is the first message received (see above).

server/server.go
kozlovic: Done

server/server_test.go
kozlovic: Done

server/server_test.go
kozlovic: Done

server/server_test.go
kozlovic: Done

ColinSullivan1:
LGTM!

PR #199, opened 2016-11-21
Updates for Go client rename, update vendor dependencies

kozlovic:


ColinSullivan1:
It looks like some close request updates were included here in vendoring - was that intended?  LGTM.

kozlovic:
@ColinSullivan1 Since I did an vendor update and the go-nats-streaming client has been already updated with new subscription.close protocol, yes, this is normal. The upcoming subscription.close support in server will therefore not have the vendor update since this will be done here.

PR #247, opened 2017-01-31
[FIXED] Sending data before SubscriptionResponse

kozlovic:
When a subscription request is received, the subscription is added
to the server and the response is sent back to the client before
starting to deliver messages to this subscription. However, in
the subscription request process, as soon as the subscription is
added but before the response is sent, if an application is sending
messages to the channel, it was possible that server would start
sending data to that subscription that had not yet fully initialized.

The issue was discovered with the node client:
https://github.com/nats-io/node-nats-streaming/issues/31

This was not affecting other Streaming client due to the use of
locking when processing the subscription request and processing
of incoming data.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -234,6 +234,7 @@ type subState struct {
 	ackTimer     *time.Timer
 	ackSub       *nats.Subscription
 	acksPending  map[uint64]int64 // key is message sequence, value is expiration time.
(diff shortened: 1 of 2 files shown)


aricart:
LGTM

StarpTech:
Awesome great job! I will try it asap.

PR #249, opened 2017-02-03
[ADDED] FileStore: option to limit use of File Descriptors

kozlovic:
Each channel corresponds to a sub-directory and each
sub-directory contain several files. The more channels are
added to the system, the more FDs are needed.

This option is a soft limit instructing the store to
try to keep the number of opened files below this limit.
It is possible that during concurrent read/write across
multiple channels the number of FDs goes above the limit.

Note that this is for the file store's FDs, not the server
as a whole. For instance, if the server embeds the NATS
server, the number of FDs for the nats-streaming-server
process could be much higher (additional FD per Streaming
client connecting to the server). If that's an issue, one
should configure the nats-streaming-server so that it
connects to a remote NATS Server.

### DIFF
--- a/stores/filestore.go
+++ b/stores/filestore.go
@@ -91,6 +91,9 @@ const (
 
 	// This is the default amount of time a message is cached.
 	defaultCacheTTL = time.Second
+
+	// defaultFileFlags are the default file flags used when opening a file
+	defaultFileFlags = os.O_RDWR | os.O_CREATE | os.O_APPEND
 )
 
 // FileStoreOption is a function on the options for a File Store
@@ -145,6 +148,10 @@ type FileStoreOptions struct {
 	// extension). It is the responsability of the script to move/remove
 	// those files.
 	SliceArchiveScript string
+
+	// FileDescriptorsLimit is a soft limit hinting at FileStore to try to
+	// limit the number of concurrent opened files to that limit.
+	FileDescriptorsLimit int64
 }
 
 // DefaultFileStoreOptions defines the default options for a File Store.
@@ -248,6 +255,15 @@ func SliceConfig(maxMsgs int, maxBytes int64, maxAge time.Duration, script strin
 	}
 }
 
+// FileDescriptorsLimit is a soft limit hinting at FileStore to try to
+// limit the number of concurrent opened files to that limit.
+func FileDescriptorsLimit(limit int64) FileStoreOption {
+	return func(o *FileStoreOptions) error {
+		o.FileDescriptorsLimit = limit
+		return nil
+	}
+}
+
 // AllOptions is a convenient option to pass all options from a FileStoreOptions
 // structure to the constructor.
 func AllOptions(opts *FileStoreOptions) FileStoreOption {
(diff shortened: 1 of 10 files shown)


stores/filestore.go
```diff
@@ -289,12 +305,49 @@ const (
...
+}
+
+type filesManager struct {
+	sync.Mutex
+	opened  int64
+	limit   int64
+	rootDir string
+	files   map[fileID]*file
+	nextID  fileID
+	closed  bool
```
kozlovic: done

stores/filestore.go
```diff
@@ -405,20 +456,20 @@ type FileMsgStore struct {
...
-	idxFile      *os.File
+	fm           *filesManager // shortcut to ms.fs.fm
+	hasFDsLimit  bool          // shortcut to ms.fstore.opts.FileDescriptorsLimit > 0
 	bw           *bufferedWriter
 	writer       io.Writer // this is `bw.buf` or `file` depending if buffer writer is used or not
 	files        map[int]*fileSlice
-	currSlice    *fileSlice
-	rootDir      string
+	writeSlice   *fileSlice
+	channel      string
```
kozlovic: done

stores/filestore.go
```diff
@@ -665,6 +715,286 @@ func (w *bufferedWriter) checkShrinkRequest() {
...
+	return fm
+}
+
+// closeUnusedFiles cloes files that are opened and not currently in-use.
+// Since the number of opened files is a soft limit, and if this function
+// is unable to close any file, the caller will still attempt to create/open
+// the requested file. If the system's file descriptor limit is reached,
+// opening the file will fail and that error will be returned to the caller.
+// Lock is required on entry.
+func (fm *filesManager) _closeUnusedFiles(idToSkip fileID) {
```
kozlovic: Yes, you got the meaning. But it is not a new convention that I would impose on the group. I would also welcome a better way to do this (without making `filesManager` public).

stores/filestore.go
```diff
@@ -665,6 +715,286 @@ func (w *bufferedWriter) checkShrinkRequest() {
...
+		}
+	}
+	if !wasOpened && !wasClosed {
+		return false, fmt.Errorf("file %q is still probably locked", file.name)
+	}
+	return wasOpened, nil
+}
+
+// remove a file from the list of files. The initial state must be either `fileOpened`
+// or `fileClosed`. This call will loop until it can switch the file's state from
```
kozlovic: This function is calling the function above. So yes, it will loop for a certain number of attempts if the file is not in the state it should be.

stores/filestore.go
```diff
@@ -2041,19 +2468,16 @@ func (ms *FileMsgStore) backgroundTasks() {
...
-							slice.file = nil
-						}
-						if slice.idxFile != nil {
-							slice.idxFile.Close()
-							slice.idxFile = nil
-						}
-						opened--
-					}
+				// If no FD limit and this is the write slice, skip.
+				if !ms.hasFDsLimit && slice == ms.writeSlice {
```
kozlovic: The line you refer to optimizes the fact that there is no need to notify (through the setting of the atomic variable) the background task that slices should be checked if the given slice is the write slice. The code here is simply saying that when check is happening, we should not close the writeSlice if there is no FD limit.

stores/filestore.go
```diff
@@ -441,25 +492,24 @@ var (
...
-			mode |= m
-		}
-	}
-
 	// Check if file already exists
 	if s, err := os.Stat(fileName); s != nil && err == nil {
 		checkVersion = true
 	}
-	file, err := os.OpenFile(fileName, mode, 0666)
+	file, err := os.OpenFile(fileName, flags, 0666)
```
kozlovic: You mean 0600? You would want the user to be able to read/write the file.. I would suggest that we take that to a different PR and likely have a `FileStoreOption` that allows one to specify the flag they want?

ColinSullivan1: Yep, flipped them around.  User = r/w, group = r, other = nothing.  A different PR, allowing the user to supply a mask sounds good here.

stores/filestore.go
ColinSullivan1: No such option I'm aware of without refactoring into a new module.  IMO they are fine without the `_`, but just an opinion.

kozlovic:
@ColinSullivan1 I have removed the `_` prefix from the "private" functions and squashed the other update based on comment to a single commit for the merge. Note that comments to squashed commits are still avail in GitHub (clicking in `Show outdated` in places where there was a comment that is no longer relevant due to the squash). Thanks for the review!

PR #253, opened 2017-02-09
[UPDATED] Schedule subscription start to go routine

kozlovic:
Processing of a new subscription request was followed by sending
available messages to this subscription. Since sub's MaxInflight
is set by user and can be quite big, it could delay the return
of this function, effectively delaying other new subscriptions
requests. This PR schedules the send of pending or avail messages
to another go routine.


kozlovic:
I have experimented with a pool of `processSubscriptionsStart()` go-routines and "performance" was actually worse. So keeping a single go-routine for now.

Issue #263, opened 2017-03-07
Shared Storage Fault Tolerance


petemiron:
## Goals:

1. Provide a durable messaging solution with server fault tolerance.

## Requirements

1. This solution assumes a FileStore NATS Streaming configuration with a shared, replicated set of physical disks, logically mounted by multiple servers (e.g. an NFS Mount (Gluster in Google Cloud or EFS in Amazon). MemStore can not be used in the Shared Storage Fault Tolerant solution.
1. A Group of NATS Streaming Servers must be configured so that only one NATS Streaming Server is a Primary Node writing to the data store.
1. NATS Streaming Server data shall be redundant across multiple disks in multiple enclosures (separate disk, power, network).
1. All state (messages, clients, subscriptions) updates must be handled by a single Primary Node. The Primary Node must prevent multiple writers and have a method to lock the data files for a single writer -- possibly using NFS locking.
1. NATS Clients must seamlessly reconnect to another server if they were on the failed unresponsive Primary Node.

sfrooster:
What's the expected performance for this solution?  I know that's difficult to answer, but I'd recommend (and personally appreciate, even help contribute to) a MongoDB implementation along side what is proposed above.  This has lots of benefits (one being price) and, I believe, has the possibility to be equally if not more interesting to the community.  I could also be thinking only of myself - hard to tell :-)

kozlovic:
@sfrooster The storage is independent of this feature. The goal here is to have servers that are ready to take over when the active process failed. The storage is shared, it can be file store (the only persistent store that we have at the moment) or a database if an implementation becomes available. The behavior will remain the same: only one server will write/read from the storage.

petemiron:
Removed quorum requirement, as solution for fault tolerance should be able to be started with a single instance.

(removed: 1. An administrator must be able to configure the target group size. Quorum will be determined by (Group Size / 2) +1 and the cluster will not begin operations until Quorum is met.)

joeblew99:
The current design assumes a NAS or SAN backing the NATS streaming server ?

kozlovic:
@joeblew99 In Fault Tolerance mode, yes. Data is not replicated, all servers point to the same shared storage. There is only one active server accessing the store and servicing clients. Standby servers are just there to take over should the active server fail.

A bit of documentation can be found [here](https://github.com/nats-io/nats-streaming-server#fault-tolerance) and [here](https://github.com/nats-io/nats-streaming-server#shared-state).

[...]

PR #265, opened 2017-03-10
Add the possibility to disallow unknown channels to avoid DoS

maxatome:
Until an authorization framework comes up for limiting subscribing (à
la nats), add a store_limits parameter: unknown_channels_disallowed.
When set to true, only channels listed in store_limits.channels hash
are allowed to be subscribed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,15 @@
 language: go
 sudo: false
 go:
-- 1.6.4
-- 1.7.4
+- 1.7.5
+- 1.8
 install:
 - go get -t ./...
 - go get github.com/nats-io/gnatsd
 - go get github.com/mattn/goveralls
 - go get github.com/wadey/gocovmerge
-- go get honnef.co/go/tools/cmd/staticcheck
-- go get honnef.co/go/tools/cmd/gosimple
+- go get -u honnef.co/go/tools/cmd/staticcheck
+- go get -u honnef.co/go/tools/cmd/gosimple
 before_script:
 - EXCLUDE_VENDOR=$(go list ./... | grep -v "/vendor/")
 - go build
--- a/server/conf.go
+++ b/server/conf.go
@@ -176,6 +176,11 @@ func parseStoreLimits(itf interface{}, opts *Options) error {
 			if err := parsePerChannelLimits(v, opts); err != nil {
 				return err
 			}
+		case "unknown_channels_disallowed":
+			if err := checkType(k, reflect.Bool, v); err != nil {
+				return err
+			}
+			opts.UnknownChannelsDisallowed = v.(bool)
 		default:
 			// Check for the global limits (MaxMsgs, MaxBytes, etc..)
 			if err := parseChannelLimits(&opts.ChannelLimits, k, name, v); err != nil {
--- a/server/conf_test.go
+++ b/server/conf_test.go
@@ -258,6 +258,7 @@ func TestParseWrongTypes(t *testing.T) {
 	expectFailureFor(t, "store_limits:{max_age:false}", wrongTypeErr)
 	expectFailureFor(t, "store_limits:{max_age:\"foo\"}", wrongTimeErr)
 	expectFailureFor(t, "store_limits:{max_subs:false}", wrongTypeErr)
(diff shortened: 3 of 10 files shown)


kozlovic:
Hello Maxime!
Thank you for this PR: this is something we were planning to do actually for a possible FT with partitioning of channels among groups of FT servers. 

That being said, I don't think the PR could be merged as-is: my guess, looking at the code, it that all messages - regardless of their channel - would be rejected if you enable the new parameter. The reason I say that is that I see that you added code in the server's lookup and if the option is enabled, you simply reject. I don't see where you would do the lookup to check if the channel is known or not.
I think that there are 2 ways in solving that: move the check in the store implementation, which right now would be in `stores/common.go` and is shared by memory and file stores, but would maybe need to be replicated if we were to add new stores, or do the check in the server itself, but it means that we should make a copy of the PerChannel map for the server's use.

As I said, we may need to do that shortly, so it is up to you to see if you still want to experiment.
Note that we have just recently published a contributions guidelines that you can find [here](http://nats.io/documentation/CONTRIBUTING/). Also, in very near future, you should have a pull request template opening when trying to create a PR. It would look like [this](https://github.com/nats-io/gnatsd/blob/master/.github/PULL_REQUEST_TEMPLATE.md). For instance, your PR lack some testing and code coverage. Adding a test that checks that some channels are authorized while others are not would be a minimum and would have show that the PR is not working as expected.

We appreciate you trying to contribute and see the value of adding such feature. Thanks again.

maxatome:
Hello,
You are right. When I did my tests, all channels have been already created by the server, that's why it worked :(

I have a new implementation that seems to work using a third way:
- add a `IsPredeclaredChannel(channel string) bool` to the `stores.Store` interface
- implement it for `genericStore` testing whether `gs.limits.PerChannel[channel]` exists or not
- still testing in `StanServer.lookupOrCreateChannel()` but using `s.store.IsPredeclaredChannel()` in addition to `s.opts.UnknownChannelsDisallowed`

This way we can avoid copying `PerChannel` map, is'nt it? Perhaps I miss something.

I pushed these modifications in my branch, I have to complete the unit tests before pull-requesting again following your guidelines. Tell me whether you are interested or not.

Thanks for your work.

kozlovic:
@maxatome 

I see value in what you are trying to do, and really think that we will need something similar for FT with partitioning. But then, I think I would need to have the info readily available in the server anyway. The problem with the approach you are taking is that it requires a change in the Store interface, which some users have already some store implementations of their own. That would be fine to change the interface (I am actually doing just that in #263), but if I soon need to have access to the list of channels in the server code itself, it would be not ideal to rip it off from the store interface soon after it was added. The other drawback of having the check in the store is, as I said, needing to possibly replicate the check in all store implementations: as you can see, right now it is in the common store implementation, but not all stores may inherit from the generic store.

So in short, I would say that for now you can proceed on your fork, but I would not consider merging until I am progressing a bit more in the channels partitioning idea that will follow the FT work for #263. Is that fair?

As a side note, your latests changes in the PR failed because of the tool we use to do some checks was temporarily broken in that they were using Go 1.8 specific code, which they reverted. I have updated travis and fixed some errors reported by those tools. When you do add to your branch, I would recommend that you first rebase with master (I have merged #267).

Thanks again for your contribution!

maxatome:
@kozlovic I understand, that's why I asked you whether you are interested or not. No problem :)
I developed this feature because I need it to avoid service abuse.

Note that we can achieve the same result without modifying any code, by reducing the store_limits.max_channels to the number of store_limits.channels items and then immediately subscribe one time to each of them, with the difference that the error message received by the client subscribing to a wrong channel is a bit confusing (too many channels), but it's a detail.

Thanks for your time, I let you close this pull request.

kozlovic:
@maxatome Thank you for your understanding. As I said, I may need to use the list of channels on startup to have internal subscriptions on these subjects only (for channels partitioning on server servers with same cluster ID). If that is the case, servers would only receive messages on those channels anyway, so there would not be a need to check. Since I would need/use the list in the server and not the store, even without the partitioning feature, it would then be easy for the server to check in `lookupOrCreateChannel`.

So I am closing this PR now but will be sure to reference it/mention it when implementing the other approach.

Thanks again for your contribution and using NATS Streaming!

PR #268, opened 2017-03-15
[ADDED] Fault Tolerance based on Shared Storage

### LINKED ISSUE #263: Shared Storage Fault Tolerance

kozlovic:
This PR adds Fault Tolerance for a group of streaming servers, using a shared storage.

Having several commits intentionally to ease the review. For instance first commit is just adding graft as vendor and does not need to be reviewed. 

 - [X] Link to issue
 - [X] Documentation added (if applicable)
 - [X] Tests added
 - [X] Branch rebased on top of current master (`git pull --rebase origin master`)
 - [ ] Changes squashed to a single commit (described [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
 - [X] Build is green in Travis CI
 - [X] You have certified that the contribution is your original work and that you license the work to the project under the [MIT license](https://github.com/nats-io/gnatsd/blob/master/LICENSE)

Resolves #263

### Changes proposed in this pull request: 

 - Updated the Store Interface to specified recovery and a way for a server to get an exclusive lock on a store. 
 - Added file locking based on advisory locks for unix'es and special windows open file call.
 - When in FT mode, server starts as a standby until it is elected active, in which case it recovers the state and starts servicing clients.
- Documentation in README.md about FT but also a pass at basic Streaming concepts that were lacking.
 
/cc @nats-io/core


server/conf.go
```diff
@@ -114,6 +114,21 @@ func ProcessConfigFile(configFile string, opts *Options) error {
 				return err
 			}
 			opts.AckSubsPoolSize = int(v.(int64))
+		case "ft_group", "ft_group_name":
```
kozlovic: No, this was taken from NATS server code. It is basically any parameter name that I think would make it easier to use or remember. In some NATS Server you will even find some with typos, which is on purpose: Derek added the ones that he usually get wrong ;-)

server/ft.go
```diff
@@ -0,0 +1,187 @@
...
+				if sc.To == graft.LEADER {
+					break waitForLeaderLoop
+				}
+			case err := <-errChan:
+				node.Close()
+				return fmt.Errorf("ft: error: %v", err)
+			}
+		}
+	}
+	Noticef("STAN: Server elected leader, getting store exclusive lock...")
```
kozlovic: Thanks, will address that.

nats-streaming-server.go
```diff
@@ -195,6 +198,9 @@ func parseFlags() (*stand.Options, *natsd.Options) {
 	flag.Int64("file_fds_limit", stores.DefaultFileStoreOptions.FileDescriptorsLimit, "FileStoreOpts.FileDescriptorsLimit")
 	flag.Int("io_batch_size", stand.DefaultIOBatchSize, "IOBatchSize")
 	flag.Int64("io_sleep_time", stand.DefaultIOSleepTime, "IOSleepTime")
+	flag.String("ft_group", "", "FTGroupName")
+	flag.Int("ft_quorum", 1, "FTQuorum")
```
kozlovic: I started with group size but then you have to do the math to know that with (our) raft, quorum will be `group_size/2+1`. So I took the opposite approach where the user specifies the quorum, which is the number of servers required for the election. They can be running more servers if they want.

server/ft_test.go
```diff
@@ -0,0 +1,687 @@
...
+	if err == nil {
+		s.Shutdown()
+		t.Fatal("Expected server to fail due to negative group size")
+	}
+	// Try with size 0 too.
+	opts.FTQuorum = 0
+	s, err = RunServerWithOpts(opts, nil)
+	if err == nil {
+		s.Shutdown()
+		t.Fatal("Expected server to fail due to negative group size")
```
kozlovic: Ok

README.md
```diff
@@ -11,15 +11,269 @@ NATS Streaming provides the following high-level feature set.
...
+## Relation to NATS
+
+NATS Streaming Server by default embeds a [NATS](https://github.com/nats-io/gnatsd) server. That is, the Streaming server is not a server per-se, but instead, a client to a NATS Server.<br>
+It means that Streaming clients are not directly connected to the streaming server, but instead communicate with the streaming server *through* NATS Server.
+
+This detail is important when it comes to Streaming clients connections to the Streaming server. Indeed, since there is no direct
+connection, the server knows if a client is connected based on heartbeats.
+
+***It is therefore strongly recommended for clients to close their connection when the application exit, otherwise the server
+will consider these clients connected (sending data, etc...) until it detects the miss of hearbeats.***
```
kozlovic: Ok

README.md
```diff
@@ -11,15 +11,269 @@ NATS Streaming provides the following high-level feature set.
...
+### Message Log
+
+You can view a message log as a ring buffer. Messages are appended to the end of the log. If a limit is set gobally for all channels, or specifically for this channel, when the limit is reached, older messages are removed to make room for the new ones.
+
+But except for the administrative size/age limit set for a message log, messages are not removed due to consumers consuming them.
+In fact, messages are stored regardless of the presence of subscriptions on that channel.
+
+### Subscriptions
+
+A client creates a subscription on a given channel. Remember, there is no support for wildcard, so a subscription is really tied to
```
kozlovic: Ok

server/ft.go
```diff
@@ -0,0 +1,187 @@
...
+	s.wg.Add(1)
+	s.mu.Unlock()
+	Noticef("STAN: Server confirmed leader")
+	go func() {
+		// On shutdown, server is waiting on s.wg, so release when
+		// we exit this function.
+		defer s.wg.Done()
+		// We are going to care only about state changes here, and not the
+		// error channel. The reason is that for a leader, the only case
+		// where we would get an error is writting the state, which would
```
kozlovic: Ok

server/ft.go
```diff
@@ -0,0 +1,187 @@
...
+// from previous instance to interfere with new one.
+func (s *StanServer) ftCreateRAFTNode(opts *nats.Options) (*graft.Node, chan graft.StateChange, chan error, error) {
+	// Loop until we can get a valid connection...
+	var (
+		err error
+		rpc *graft.NatsRpcDriver
+	)
+	// We need to have a NATS RPC in order for RAFT to work,
+	// so try until we succeed.
+	printCount := 0
```
kozlovic: Can rename to attempts indeed.

server/ft.go
```diff
@@ -0,0 +1,187 @@
...
+		rpc, err = graft.NewNatsRpc(opts)
+		if err != nil {
+			if printCount%10 == 0 {
+				Errorf("STAN: ft: creating RPC failed (%v), trying again...", err)
+			}
+			printCount++
+			select {
+			case <-s.ftQuit:
+				return nil, nil, nil, nil
+			case <-time.After(500 * time.Millisecond):
```
kozlovic: RAFT will not work without the RPC in place, so I want to try as often as possible - not too crazy - but not delay more than needed.

nats-streaming-server.go
ColinSullivan1: So long as it is clear to a typical user that this is the minimum # of servers, sounds good here.

kozlovic: In the command line output, the server's Options description and README, I document this parameter as: `Number of servers needed in order to elect a leader`. Let's see if Peter objects too and we will see if we revert to specify group size instead (again, my problem with that is that it sounds like you define a rigid group and yet, you don't know how many servers will be needed for the election unless you do the math: size/2+1).

kozlovic:
@ColinSullivan1 I hope I addressed most of your comments. Specifically, let me know if you find the main ft function simpler to read this way.

server/conf.go
ColinSullivan1: Sounds good then.

kozlovic:
@ColinSullivan1 @petemiron I am reconsidering the use of RAFT for that. I may send another PR as an alternative approach. @petemiron if you have not started the review, maybe hold off on that.

ColinSullivan1:
@kozlovic , I can see RAFT being a bit heavyweight alongside file locking.  Perhaps encapsulating raft in an API that is friendlier to FT usage here could be another option.

kozlovic:
Closing in favor of #270

PR #269, opened 2017-03-17
Added BackoffTimeCheck utility

kozlovic:
In situations where an error/trace is printed but there is a
possibility that the print is too frequent, which could cause the
log to quickly fill up, this utility should be used.

You first create a BackoffPrint object giving a minimum and maximum
frequency and a factor at which the frequency is increased.

An Ok() function tells you if it is ok to issue the print statement.
When returning true, the object sets its next print target which
causes subsequent calls to Ok() to return false if they are made
too soon.

The frequency allowed to print is increased by the given factor
anytime the function returns Ok(). It is capped by the max frequency
set when creating the object.

After reaching the max frequency, if a call to Ok() is made after
more than 3x the max frequency since the last call that returned
true, then the call returns Ok() but is also auto-reset.

There is also a Reset() API to explicitly reset the object
to an initial state.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -6,6 +6,7 @@ import (
 	"encoding/binary"
 	"io"
 	"os"
+	"time"
 )
 
 // ByteOrder specifies how to convert byte sequences into 16-, 32-, or 64-bit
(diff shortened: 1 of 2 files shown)


util/util.go
```diff
@@ -16,6 +17,65 @@ func init() {
...
+// and the allowed frequency) has elapsed.
+// When at the maximum frequency, if this call is made after a delay at least
+// equal to 3x the max frequency (or in other words, 2x after what was the target
+// for the next print), then the object is auto-reset.
+func (bp *BackoffPrint) Ok() bool {
+	if bp.nextPrint.IsZero() {
+		bp.nextPrint = time.Now().Add(bp.minFrequency)
+		return true
+	}
+	if time.Now().Before(bp.nextPrint) {
```
ColinSullivan1: time.Now() is used a few times in this function; you can call it once and assign a var to it.

util/util.go
```diff
@@ -16,6 +17,65 @@ func init() {
...
+	maxFrequency time.Duration
+	factor       int
+}
+
+// NewBackoffPrint creates an instance of BackoffPrint.
+// The `minFrequency` indicates how frequently BackoffPrint.Ok() can return true.
+// When Ok() returns true, the allowed frequency is multiplied by `factor`. The
+// resulting frequency is capped by `maxFrequency`.
+func NewBackoffPrint(minFrequency time.Duration, factor int, maxFrequency time.Duration) *BackoffPrint {
+	return &BackoffPrint{
```
ColinSullivan1: I'd suggest some error checking here - it could make the life of someone who misconfigures this much easier.

util/util.go
kozlovic: Actually no, since there are returns. Each time time.Now() is invoked, it is once per code path.
Unless I change `time.Since()` to do something like `now.Sub(bp.nextPrint)`, which I could.

util/util.go
kozlovic: This is for internal use, though, but ok, by checking you mean max>=min, min>=0, factor>=1, things like that?

util/util.go
```diff
@@ -16,6 +17,65 @@ func init() {
 	ByteOrder = binary.LittleEndian
 }
 
+// BackoffPrint allows to print a statement but not too often.
```
kozlovic: Good idea.

util/util.go
kozlovic: Yes, changing that too.

PR #270, opened 2017-03-18
[ADDED] Shared Storage Fault Tolerance

### LINKED ISSUE #263: Shared Storage Fault Tolerance

kozlovic:
This PR is a second take at FT, without the use of RAFT as in #268.

 - [X] Link to issue
 - [X] Documentation added (if applicable)
 - [X] Tests added
 - [X] Branch rebased on top of current master (`git pull --rebase origin master`)
 - [ ] Changes squashed to a single commit (described [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
 - [X] Build is green in Travis CI
 - [X] You have certified that the contribution is your original work and that you license the work to the project under the [MIT license](https://github.com/nats-io/gnatsd/blob/master/LICENSE)

Resolves #263

### Changes proposed in this pull request: 

 - Updated the Store Interface to specified recovery and a way for a server to get an exclusive lock on a store. 
 - Added file locking based on advisory locks for unix'es and special windows open file call.
 - Added a DelegateStore that can be used to create mock stores in order to override some store APIs for test purposes.
 - When in FT mode, server starts as a standby until it is elected active, in which case it recovers the state and starts servicing clients.
- Documentation in README.md about FT but also a pass at basic Streaming concepts that were lacking.
 
/cc @nats-io/core

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,246 @@ NATS Streaming provides the following high-level feature set.
 - Replay/Restart
 - Last Value Semantics
 
-### Getting Started
+# Table of Contents
+
+- [Important Changes](#important-changes)
+- [Concepts](#concepts)
+    * [Relation to NATS](#relation-to-nats)
+    * [Clients Connections](#clients-connections)
+    * [Channels](#channels)
+        * [Message Log](#message-log)
+        * [Subscriptions](#subscriptions)
+            * [Regular](#regular)
+            * [Durable](#durable)
+            * [Queue Group](#queue-group)
+            * [Redelivery](#redelivery)
+    * [Store Interface](#store-interface)
+    * [Clustering](#clustering)
+    * [Fault Tolerance](#fault-tolerance)
+        * [Active Server](#active-server)
+        * [Standby Servers](#standby-servers)
+        * [Shared State](#shared-state)
+        * [Failover](#failover)
+- [Getting Started](#getting-started)
+    * [Building](#building)
+    * [Running](#running)
+- [Configuring](#configuring)
+    * [Command line arguments](#command-line-arguments)
+    * [Configuration file](#configuration-file)
+    * [Store Limits](#store-limits)
+        * [Limits inheritance](#limits-inheritance)
+    * [Securing](#securing)
+        * [Authorization](#authorization)
+        * [TLS](#tls)
+    * [Persistence](#persistence)
+        * [File Store](#file-store)
+            * [File Store Options](#file-store-options)
+- [Clients](#clients)
+- [Licence](#license)
+
+# Important Changes
+
+## Version `0.4.0`
+
+The Store interface was updated. There are 2 news APIs:
+
+* `Recover()`: The recovery of persistent state was previously done in the constructor of the store implementation.<br>
+It is now separate and specified with this API. The server will first instantiate the store, in
+which some initialization or checks can be made.<br>
+If no error is reported, the server will then proceed with calling `Recover()`, which will returned the recovered state.<br>
+* `GetExclusiveLock()`: In Fault Tolerance mode, when a server is elected leader, it will attempt to get an exclusive
+lock to the shared storage before proceeding.<br>
+
+Check the [Store interface](https://github.com/nats-io/nats-streaming-server/blob/master/stores/store.go) for more information.
+
+# Concepts
+
+## Relation to NATS
+
+NATS Streaming Server by default embeds a [NATS](https://github.com/nats-io/gnatsd) server. That is, the Streaming server is not a server per-se, but instead, a client to a NATS Server.<br>
+It means that Streaming clients are not directly connected to the streaming server, but instead communicate with the streaming server *through* NATS Server.
+
+This detail is important when it comes to Streaming clients connections to the Streaming server. Indeed, since there is no direct
+connection, the server knows if a client is connected based on heartbeats.
+
+***It is therefore strongly recommended for clients to close their connection when the application exit, otherwise the server
+will consider these clients connected (sending data, etc...) until it detects missing hearbeats.***
+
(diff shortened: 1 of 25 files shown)


kozlovic:
@petemiron I realize now that we have in the requirements a mention of Quorum (which implies RAFT). Not sure if this is required. If so, then this PR won't work.
```
3. An administrator must be able to configure the target group size.
Quorum will be determined by (Group Size / 2) +1 and the cluster will not begin 
operations until Quorum is met.
```

ColinSullivan1:
@kozlovic @petemiron Regarding the requirements, IMO a quorum isn't required here.

petemiron:
@kozlovic @ColinSullivan1 agreed. requirement removed.

server/ft.go
```diff
@@ -0,0 +1,214 @@
...
+		case <-s.ftHBCh:
+			// go back to the beginning of the for loop
+			continue
+		case <-time.After(ftHBMissedInterval):
+			// try to lock the store
+		}
+		locked, err := s.ftGetStoreLock()
+		if err != nil {
+			// TODO: This means that we got an error not related to locking.
+			// Since we are in standby, should just keep trying? It could
```
ColinSullivan1: IMO keep trying, could be a temporary file system issue.  Do you think there should be the concept of a fatal error vs an error that warrants a retry?

kozlovic: My concern with retrying was this: If the user mistakenly gives the bad path, etc.. then the server will keep trying and user may not know of the error. That being said, not sure if (or how safe) it is to detect an transient error to a configuration error. But agreed that we may want to try for ever here.

kozlovic: Remember that `store.GetExclusiveLock()` is a store API, so depending on the store implementation, Permission or Path doesn't exist may not make sense. We should then specify at the Store.GetExclusiveLock() the behavior, for instance return error if failing the get the lock is considered fatal as opposed to maybe temporary? If that's the case, the changes would be in this PR in filestore and here I would indeed return error if `err != nil`.

kozlovic:
@petemiron Updated the PR with some changes we discussed.

README.md
```diff
@@ -11,15 +11,282 @@ NATS Streaming provides the following high-level feature set.
...
+
+## Clustering
+
+NATS Streaming Server does not support clustering at this time. The confusion is that it can run with a cluster of NATS Servers,
+but as previously explained, the NATS Servers are the backbone of the streaming systems. So it is possible to run a single
+streaming server attached to a cluster of NATS Servers and streaming clients attached to any of the NATS Server in the NATS cluster.
+
+Streaming clustering with full replication of data across a set of streaming servers is planned for later this year.
+
+## Fault Tolerance
```
petemiron: Can we add a diagram here? or maybe we reference another page.

util/util.go
```diff
@@ -4,12 +4,22 @@ package util
...
+	"errors"
 	"fmt"
 	"io"
-	"os"
 	"time"
 )
 
+// ErrUnableToLockNow is used to indicate that a lock cannot be
+// immediately acquired.
+var ErrUnableToLockNow = errors.New("unable to acquire the lock at the moment")
```
kozlovic: Then I can't use a generic error since the name of the file is given to the `CreateLockFile`. Also, the caller in this case is `FileStore.GetExclusiveLock()` which when the lock cannot be acquired because it is already locked returns `false`, but no error. So I would not even have a way to report the file name. Moreover, `Store.GetExclusiveLock()` will not necessarily be file related if/when we have other type of stores.

stores/filestore.go
```diff
@@ -1139,15 +1149,35 @@ func NewFileStore(rootDir string, limits *StoreLimits, options ...FileStoreOptio
...
+}
+
+// GetExclusiveLock implements the Store interface
+func (fs *FileStore) GetExclusiveLock() (bool, error) {
+	fs.Lock()
+	defer fs.Unlock()
+	if fs.lockFile != nil {
+		return true, nil
+	}
+	f, err := util.CreateLockFile(filepath.Join(fs.fm.rootDir, "ft.lck"))
```
kozlovic: I should have made this a constant at the very least (especially that I use that name in a test). Will change that.

README.md
```diff
@@ -11,15 +11,282 @@ NATS Streaming provides the following high-level feature set.
...
+        * [TLS](#tls)
+    * [Persistence](#persistence)
+        * [File Store](#file-store)
+            * [File Store Options](#file-store-options)
+- [Clients](#clients)
+- [Licence](#license)
+
+# Important Changes
+
+## Version `0.4.0`
```
kozlovic: This is not a change log per-se. Most of the time there will be no addition to that section. This is just when we either break backward compatibility or introduce/change Store interface for people that have done (or planning to do) store implementations.

README.md
kozlovic: I will see what I can do ;-)

Issue #62, opened 2016-04-29
Cleanup Logging


ColinSullivan1:
Remove "STAN:" from the individual log lines to instead prefix in the logger.

petemiron:
Fixed in PR #274

PR #274, opened 2017-03-31
[ADDED] Cleanup Logging

petemiron:
Closes Issue #62 .

Changed prefix to STREAM from STAN.

Please double check unit test in server/log_test.go. It's a little weird due to the empty string case.

Also, to confirm, should we lowercase the first word in all log messages?

### DIFF
--- a/server/log_test.go
+++ b/server/log_test.go
@@ -147,8 +147,12 @@ func TestLogOutput(t *testing.T) {
 	d := &dummyLogger{}
 
 	checkLogger := func(output string) {
-		if d.msg != output {
-			t.Fatalf("Unexpected logger message: %v", d.msg)
+		checkMsg := output
+		if checkMsg != "" {
+			checkMsg =  LogPrefix + checkMsg
+		}
+		if checkMsg != d.msg {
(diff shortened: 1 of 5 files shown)


server/log_test.go
```diff
@@ -147,8 +147,12 @@ func TestLogOutput(t *testing.T) {
 	d := &dummyLogger{}
 
 	checkLogger := func(output string) {
-		if d.msg != output {
-			t.Fatalf("Unexpected logger message: %v", d.msg)
+		checkMsg := output
```
kozlovic: Please revert those changes. You don't need to do anything special here. I am using a dummy logger (see code above) and this is not using the normal logger. So whatever text is passed to `Noticef`, `Debugf`, etc.. in this test is set in `dummyLogger.msg`.

kozlovic:
Go says that if you create an `error` (even with `fmt.Errorf()`), the first letter should be lower case and not have period at the end. However, this is our logger and that does not apply. For instance, linters should not report those as warnings.

server/log_test.go
kozlovic: After discussion, I realize my mistake since the addition of the prefix is done before delegating to the (dummy) logger.

ColinSullivan1:
LGTM.

PR #284, opened 2017-04-05
[FIX] Fix parsing of the --file_slice_max_age parameter.

ColinSullivan1:
The `--file_slice_max_age` parameter was not being parsed as a duration.


kozlovic:
Thanks, good catch!

PR #285, opened 2017-04-05
[ADDED] Partitioning

### LINKED ISSUE #271: Support partitioning by channel
A cluster of NATS streaming servers should be able to distribute work through horizontally partitioning channels.  Each server in a cluster can service a set of channels.  This will increase performance and allow users to scale NATS streaming.

Requirements:
* Allow multiple active servers using the same cluster ID.
* A channel can be serviced by exactly one streaming server at a time.
 
Discussion points:
* Should channel assignments be predetermined, or randomized, or allow both?
* How do we enforce one channel per cluster? (_e.g. Use RAFT consensus to have a leader decide who owns a channel, then have all members update configuration files with white and black lists?_)
* Can channels be "moved"?

kozlovic:
This feature allows to have multiple server instances on the
same cluster ID but each instance handling a list of channels.
A given channel must not be defined on more than one instance.
Servers on the same cluster ID try to ensure that this is the case.

This feature also allows to prevent creation of unwanted channels.
A similar feature was proposed by @maxatome in PR #265 that was
not accepted because there were plans for FT and Channels Partitioning.

Resolves #271

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ NATS Streaming provides the following high-level feature set.
         * [Standby Servers](#standby-servers)
         * [Shared State](#shared-state)
         * [Failover](#failover)
+    * [Partitioning](#partitioning)
 - [Getting Started](#getting-started)
     * [Building](#building)
     * [Running](#running)
@@ -74,7 +75,7 @@ This detail is important when it comes to Streaming clients connections to the S
 connection, the server knows if a client is connected based on heartbeats.
 
 ***It is therefore strongly recommended for clients to close their connection when the application exit, otherwise the server
-will consider these clients connected (sending data, etc...) until it detects missing hearbeats.***
+will consider these clients connected (sending data, etc...) until it detects missing heartbeats.***
 
 The streaming server creates internal subscriptions on specific subjects to communicate with its clients and/or other servers.
 
@@ -109,7 +110,7 @@ in a message log inside this channel.
 
 ### Message Log
 
-You can view a message log as a ring buffer. Messages are appended to the end of the log. If a limit is set gobally for all channels, or specifically for this channel, when the limit is reached, older messages are removed to make room for the new ones.
+You can view a message log as a ring buffer. Messages are appended to the end of the log. If a limit is set globally for all channels, or specifically for this channel, when the limit is reached, older messages are removed to make room for the new ones.
 
 But except for the administrative size/age limit set for a message log, messages are not removed due to consumers consuming them.
 In fact, messages are stored regardless of the presence of subscriptions on that channel.
@@ -195,7 +196,7 @@ because when the server runs on Fault Tolerance mode, the storage must be shared
 using it.
 
 After instantiating the store, the server will then call `Recover()` in order to recover the persisted state. For implementations
-that do not support persitence, such as the provided `MemoryStore`, this call will simply return `nil` (without error) to
+that do not support persistence, such as the provided `MemoryStore`, this call will simply return `nil` (without error) to
 indicate that no state was recovered.
 
 The `Store` is used to add/delete clients, create/lookup channels, etc...
@@ -222,8 +223,7 @@ To minimize the single point of failure, NATS Streaming server can be run in Fau
 of servers with one acting as the active server (accessing the store) and handling all communication with clients, and all others
 acting as standby servers.
 
-To start a server in Fault Tolerance (FT) mode, you specify an FT group name. This is so that in the future, one could have
-different FT groups for servers with the same cluster ID. For now, use it to simply enable Fault Tolerance mode.
+To start a server in Fault Tolerance (FT) mode, you specify an FT group name.
 
 Here is an example on how starting 2 servers in FT mode running on the same host and embedding the NATS servers:
 
@@ -243,7 +243,7 @@ elected server fails to grab this lock because it is already locked, it will go 
 
 ### Standby servers
 
-There can be as many as you want standby servers on the same group. These servers do not access the store and do not receive any data from the streaming clients. They are just running waiting for the detectiong of the active server failure.
+There can be as many as you want standby servers on the same group. These servers do not access the store and do not receive any data from the streaming clients. They are just running waiting for the detection of the active server failure.
 
 ### Shared State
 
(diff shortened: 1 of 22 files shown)


kozlovic:
I am going so see how possible it would be to make use of wildcards for static channels configuration. Wildcards would still be rejected when publishing or subscribing to channels though.

kozlovic:
@ColinSullivan1 @petemiron I have added support for wildcards for the static channels feature. Note that I had recently rejected wildcards in config (https://github.com/nats-io/nats-streaming-server/pull/281) because until now there was no support. However, in this PR, the server again allows wildcards channels in the config, however, channels are incorrectly treated as literals for store limits. I will address that in a separate PR since this is not directly related to static channels.

README.md
```diff
@@ -264,6 +266,99 @@ It is possible that a standby trying to activate is not able to immediately acqu
 it goes back into standby mode, but if it fails to receive heartbeats from an active server, it will try again to
 acquire the store lock. The interval is random but as of now set to a bit more than a second.
 
+## Static Channels aka Partitioning
```
petemiron: I would recommend sticking with one name here. Probably something along the lines of "Channel Partitions". Maybe even simply "Partitions" and make sure to keep it consistent without the dual options. If we need to make sure people find what they're looking for when they search, we can add it into the help text.

README.md
```diff
@@ -264,6 +266,99 @@ It is possible that a standby trying to activate is not able to immediately acqu
...
+common, it will return an error to the emitting server, which will then fail to start.
+
+However, on startup, it is possible that the undelying NATS cluster is not fully formed. The server would
+not get any response for the rest of the cluster and therefore start successully and service clients.
+Anytime a Streaming server detects that a NATS server was added to the NATS cluster, it will resend its list
+of channels. It means that currently running servers may suddendly fail with a message regarding duplicate channels.
+It also means that the same channel may have been created in different servers, the same subscription accepted
+in different servers (but only one would be able to process application acknowledgements), etc...
+
+***In order to avoid issues with channels existing on several servers, it is ultimately the responsability
```
petemiron: How do I have Fault Tolerance with Partitioning?

README.md
```diff
@@ -264,6 +266,99 @@ It is possible that a standby trying to activate is not able to immediately acqu
...
+When a server starts, it sends its list of channels to all other servers on the same cluster in an attempt
+to detect duplicate channels. When a server receives this list and finds that it has a static channel in
+common, it will return an error to the emitting server, which will then fail to start.
+
+However, on startup, it is possible that the undelying NATS cluster is not fully formed. The server would
+not get any response for the rest of the cluster and therefore start successully and service clients.
+Anytime a Streaming server detects that a NATS server was added to the NATS cluster, it will resend its list
+of channels. It means that currently running servers may suddendly fail with a message regarding duplicate channels.
+It also means that the same channel may have been created in different servers, the same subscription accepted
+in different servers (but only one would be able to process application acknowledgements), etc...
```
petemiron: I'd drop the `etc...` 'in different servers, but only one server will process application acknowledgements.

README.md
```diff
@@ -491,6 +586,14 @@ ack_subs_pool_size: 10
 # Can be ft_group, ft_group_name
 ft_group: "ft"
 
+# In order to use static channels (or channels partitioning), this parameter
+# needs to be set to true and the list of channels defined in
+# store_limits/channels section. This section then serves two purposes,
+# overriding limits for a given channel or simply defining it as a static
+# channel.
+# Can be static_channels, channels_partitioning
```
petemiron: Give this one name.

README.md
```diff
@@ -264,6 +266,99 @@ It is possible that a standby trying to activate is not able to immediately acqu
...
+### Wildcards
+
+NATS Streaming does not support sending or subscribing to wildcard channels (such as `foo.*`).
+
+However, it is possible to use wildcards to define the static channels that a server can handle.
+For instance, with the following configuration:
+```
+static_channels: true
+store_limits: {
+    channels: {
```
petemiron: should channels be a map or an array?

README.md
kozlovic: What about "Partitioning" then? I would also change the `static_channels` parameter to the boolean `partitioning`?

README.md
kozlovic: I probably could have been a simple array, unfortunately that was not added in this PR and if changing now would break existing configuration files.

README.md
kozlovic: `partitioning` ok?

kozlovic:
@ColinSullivan1 @petemiron Updated the PR with a rename (removed reference of static channels).

kozlovic:
@petemiron I also added a section on how to use FT and Partitioning together.

README.md
```diff
@@ -264,6 +264,152 @@ It is possible that a standby trying to activate is not able to immediately acqu
...
+5. ...
+
+### A given channel must be defined in a single server
+
+When a server starts, it sends its list of channels to all other servers on the same cluster in an attempt
+to detect duplicate channels. When a server receives this list and finds that it has a channel in
+common, it will return an error to the emitting server, which will then fail to start.
+
+However, on startup, it is possible that the undelying NATS cluster is not fully formed. The server would
+not get any response from the rest of the cluster and therefore start successully and service clients.
```
ColinSullivan1: typo `successfully`

README.md
```diff
@@ -264,6 +264,152 @@ It is possible that a standby trying to activate is not able to immediately acqu
...
+Anytime a Streaming server detects that a NATS server was added to the NATS cluster, it will resend its list
+of channels. It means that currently running servers may suddendly fail with a message regarding duplicate channels.
```
ColinSullivan1: suddenly

README.md
```diff
@@ -264,6 +264,152 @@ It is possible that a standby trying to activate is not able to immediately acqu
...
+Having the same channel on different servers mean that a subscription would be created on all servers
```
ColinSullivan1: means

README.md
```diff
@@ -264,6 +264,152 @@ It is possible that a standby trying to activate is not able to immediately acqu
...
+handling the channel, but only one server will receive and process message acknowledgements. Other servers
+would then redeliver messages (since they would not get the acknowledgements), which would cause duplicates.
+
+***In order to avoid issues with channels existing on several servers, it is ultimately the responsability
```
ColinSullivan1: responsibility

README.md
```diff
@@ -264,6 +264,152 @@ It is possible that a standby trying to activate is not able to immediately acqu
...
+of the administrator to ensure that channels are unique.***
+
+### Fault Tolerance and Partitioning
```
ColinSullivan1: +1 for this example!

README.md
```diff
@@ -264,6 +264,152 @@ It is possible that a standby trying to activate is not able to immediately acqu
...
+    channels: {
+        bar.>: {}
+    }
+}
+```
+
+If you remember, Fault Tolerance is configured by specifying a name (`ft_group_name`). Suppose there is
+an NFS mount called `/nss/datastore` on both `host1` and `host2`.
+
+Starting an FT pair for the partion `foo` could look like this:
```
ColinSullivan1: partition

server/partitions.go
```diff
@@ -0,0 +1,343 @@
...
+	// If later the known servers change, it will be interpreted as
+	// a change in the NATS cluster, so we will send our channels list
+	// again.
+	p.updateServersMap(p.nc.DiscoveredServers())
+	// Now send our list and check if any server is complaining
+	// about having one channel in common.
+	if err := p.checkChannelsUniqueInCluster(); err != nil {
+		return err
+	}
+	p.msgsCh = make(chan *nats.Msg, 65536)
```
ColinSullivan1: This could be a constant.

server/partitions.go
```diff
@@ -0,0 +1,343 @@
...
+			return fmt.Errorf("channel %q causes conflict with channels on server %q",
+				string(resp.Data), resp.ServerID)
+		}
+	}
+}
+
+// Sends the list of channels to a known subject, possibly splitting the list
+// in several requests if it cannot fit in a single message.
+func (p *partitions) sendChannelsList(replyInbox string) error {
+	sendReq := func(channels []string) error {
```
ColinSullivan1: You could adjust this by passing the encoder (or bytes), then having calling code below to call within the loop.  It could simplify the code a bit.

server/partitions.go
```diff
@@ -0,0 +1,343 @@
...
+	end := 0
+	for end < len(p.channels) {
+		size := 0
+		tmpBuf := &bytes.Buffer{}
+		encoderTmpBuf := gob.NewEncoder(tmpBuf)
+		for end = start; end < len(p.channels); end++ {
+			encoderTmpBuf.Encode(p.channels[end])
+			size += tmpBuf.Len()
+			// Leave room for the CtrlMsg header
+			if size > int(maxPayload-50) {
```
ColinSullivan1: Another nitpick.  I do realize it is arbitrary, but an old manager we had dinged me a few times for not using sizes that weren't based on pointer/register sizes for things like these.  IMO, a good thing.  64 here?

server/server.go
```diff
@@ -1053,7 +1068,9 @@ func (s *StanServer) start(runningState State) error {
 		}
 	}
 
-	if runningState == Standalone {
+	// We don't do the check if we are running FT and/or if
+	// static channels (partitioning) is in play.
+	if runningState == Standalone && !s.opts.Partitioning {
```
ColinSullivan1: `s.partitions == nil` is used elsewhere to check for partitioning.

server/partitions.go
kozlovic: Sure thing.

server/partitions.go
kozlovic: Not sure what you mean here. This is a helper function to encode a part of the channels array and do the nats.Conn.Publish() call.

server/partitions.go
kozlovic: Changing the code a bit to create a request object with the server ID and type set and use `Size()` to get the encoded size that I use here instead of this value.

server/server.go
kozlovic: Will do.

server/partitions.go
ColinSullivan1: Rather than passing and array of channels and encoding them here, pass the encoded bytes as you traverse the channels - it'd prevent encoding twice.

kozlovic: No because in the temp buffer done outside of this function, there is one channel too many, which brought the size above the threshold. This is why I used a temporary buffer in which I encode channels (adding 1 at a time) until the overall size is greater than the limit. I therefore *need* to re-encode with the known number of channels that fit.
Regardless, I am revisiting this whole function. Will resubmit later.

PR #299, opened 2017-04-21
Use script to invoke misspell to run for all files but exclude vendor

kozlovic:


misspell.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+if [ "$(misspell -locale US . | grep -v 'vendor/')" ]; then
```
derekcollison: -error flag?

kozlovic: No, because we don't care that misspell returns an error code, we are looking at the actual output.

derekcollison:
LGTM

PR #306, opened 2017-05-02
[IMPROVED] Error returned to client for invalid client ID

kozlovic:
Report errors (as opposed to debug tracing) when something goes
wrong during the processing of the connect request.
For invalid client ID, returns an explicit error to clients.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -110,6 +110,7 @@ var (
 	ErrInvalidSub         = errors.New("stan: invalid subscription")
 	ErrInvalidClient      = errors.New("stan: clientID already registered")
 	ErrMissingClient      = errors.New("stan: clientID missing")
+	ErrInvalidClientID    = errors.New("stan: invalid clientID: only alphanumeric and `-` or `_` characters allowed")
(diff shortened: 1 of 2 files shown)


kozlovic:
@petemiron We had several cases now where users reported spending a lot of time trying to understand why their application would fail to connect with an "Invalid connection request" error, when it turns out that it was because of the characters used for the client ID. The server was rejecting it but with the standard `ErrInvalidConnReq`. This PR is sending a specific error for the client ID issue. There is an opened issue for the go-nats-streaming client (https://github.com/nats-io/go-nats-streaming/issues/133), and although I am planning on adding a comment for the API doc of `Connect()`, I believe that it is better to address it in the server since we may later change the restriction for client ID. (what I mean is that I would not want to add clientID format checking in all clients but instead let the server decide).

kozlovic:
@petemiron Regarding coverage. No, the error you are referring to is simply not happening during the tests (it would happen if there was a race between adding and removing a client. This PR does not change that. Note that in previous PR when adding signal handling (and test for that), the whole server package ended up not being reported. I have fixed that already, but that may have some effect in some reports.

PR #307, opened 2017-05-03
Fixed code coverage

kozlovic:
PR #305 attempted to fix an issue caused by one of the test
sending a SIGINT to test that server is exiting properly when
receiving that signal. My attempt to exclude those tests and run
them separately using a regex resulted in some tests actually
not running which brought the server.go code coverage down to 90%.
I am removing test that sends SIGINT, which brings the code coverage
for signal.go down to 75%.

### DIFF
--- a/scripts/cov.sh
+++ b/scripts/cov.sh
@@ -3,8 +3,7 @@
 
 rm -rf ./cov
 mkdir cov
-go test -v -covermode=count -coverprofile=./cov/server.out -run=Test[^Signal] ./server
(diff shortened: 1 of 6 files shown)


kozlovic:
Well, I know if works since I can Ctrl+C the server and see it being shutdown ;-). I could have a test that starts the server in a different process and issue a signal to that process ID, which then would not be a problem for the test suite, but that would not help for test coverage since the remote process will not be traced for coverage.

PR #308, opened 2017-05-11
32-bit issue

opoudjis:
Compilation in 32-bit architecture fails because of overflow of 0xffffffff.


kozlovic:
Thanks! Good catch! I would probably have seen this only before the next release when running the cross compile script (that builds all flavors) ;-)

PR #317, opened 2017-05-31
[ADDED] Monitoring

### LINKED ISSUE #116: Provide Monitoring
Introspection into the NATS streaming server via a monitoring interface is required.  An interface similar to that of the NATS server should suffice.

Some discussion points:
- Which data points should be exposed, and how should they be presented  e.g. `/server`, `/stores`, `/clients`?
- Can a pass through using he embedded NATS server port work, or should the NATS streaming server have it's own dedicated port?
- How will authorization/authentication look?

kozlovic:
Added some endpoints for monitoring:
- server/
- store/
- clients/
- channels/

Resolves #116

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -1,0 +1,504 @@
+// Copyright 2017 Apcera Inc. All rights reserved.
+
+package server
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"runtime"
+	"sort"
+	"strconv"
+	"time"
+
+	gnatsd "github.com/nats-io/gnatsd/server"
+	"github.com/nats-io/nats-streaming-server/stores"
+)
+
+// Routes for the monitoring pages
+const (
+	RootPath     = "/nss"
+	ServerPath   = RootPath + "/server"
+	StorePath    = RootPath + "/store"
+	ClientsPath  = RootPath + "/clients"
+	ChannelsPath = RootPath + "/channels"
+
+	defaultMonitorListLimit = 1024
+)
+
+// Serverz describes the NATS Streaming Server
+type Serverz struct {
+	ClusterID     string    `json:"cluster_id"`
+	ServerID      string    `json:"server_id"`
+	Version       string    `json:"version"`
+	GoVersion     string    `json:"go"`
+	State         string    `json:"state"`
+	Now           time.Time `json:"now"`
+	Start         time.Time `json:"start_time"`
+	Uptime        string    `json:"uptime"`
+	Clients       int       `json:"clients"`
+	Subscriptions int       `json:"subscriptions"`
+	Channels      int       `json:"channels"`
+	TotalMsgs     int       `json:"total_msgs"`
+	TotalBytes    uint64    `json:"total_bytes"`
+}
+
+// Storez describes the NATS Streaming Store
+type Storez struct {
+	ClusterID  string             `json:"cluster_id"`
+	ServerID   string             `json:"server_id"`
+	Now        time.Time          `json:"now"`
+	Type       string             `json:"type"`
+	Limits     stores.StoreLimits `json:"limits"`
+	TotalMsgs  int                `json:"total_msgs"`
+	TotalBytes uint64             `json:"total_bytes"`
+}
+
+// Clientsz lists the client connections
+type Clientsz struct {
+	ClusterID string     `json:"cluster_id"`
+	ServerID  string     `json:"server_id"`
+	Now       time.Time  `json:"now"`
+	Offset    int        `json:"offset"`
+	Limit     int        `json:"limit"`
+	Count     int        `json:"count"`
+	Total     int        `json:"total"`
+	Clients   []*Clientz `json:"clients"`
+}
+
+// Clientz describes a NATS Streaming Client connection
+type Clientz struct {
+	ID            string                      `json:"id"`
+	HBInbox       string                      `json:"hb_inbox"`
+	Subscriptions map[string][]*Subscriptionz `json:"subscriptions,omitempty"`
+}
+
+// Channelsz lists the name of all NATS Streaming Channelsz
+type Channelsz struct {
+	ClusterID string    `json:"cluster_id"`
+	ServerID  string    `json:"server_id"`
+	Now       time.Time `json:"now"`
+	Offset    int       `json:"offset"`
+	Limit     int       `json:"limit"`
+	Count     int       `json:"count"`
+	Total     int       `json:"total"`
+	Channels  []string  `json:"channels"`
+}
+
+// ChannelsWithSubsz list detailed channels along with subscriptions on each channel
+type ChannelsWithSubsz struct {
+	ClusterID string      `json:"cluster_id"`
+	ServerID  string      `json:"server_id"`
+	Now       time.Time   `json:"now"`
+	Offset    int         `json:"offset"`
+	Limit     int         `json:"limit"`
+	Count     int         `json:"count"`
+	Total     int         `json:"total"`
+	Channels  []*Channelz `json:"channels"`
+}
+
+// Channelz describes a NATS Streaming Channel
+type Channelz struct {
+	Name          string           `json:"name"`
+	Msgs          int              `json:"msgs"`
+	Bytes         uint64           `json:"bytes"`
+	FirstSeq      uint64           `json:"first_seq"`
+	LastSeq       uint64           `json:"last_seq"`
+	Subscriptions []*Subscriptionz `json:"subscriptions,omitempty"`
+}
+
+// Subscriptionz describes a NATS Streaming Subscription
+type Subscriptionz struct {
+	Inbox        string `json:"inbox"`
+	AckInbox     string `json:"ack_inbox"`
+	DurableName  string `json:"durable_name,omitempty"`
+	QueueName    string `json:"queue_name,omitempty"`
+	IsDurable    bool   `json:"is_durable"`
+	MaxInflight  int    `json:"max_inflight"`
+	AckWait      int    `json:"ack_wait"`
+	LastSent     uint64 `json:"last_sent"`
+	PendingCount int    `json:"pending_count"`
+	IsStalled    bool   `json:"is_stalled"`
+}
+
+func (s *StanServer) startMonitoring(nOpts *gnatsd.Options) error {
+	var hh http.Handler
+	// If we are connecting to remote NATS Server, we start our own
+	// HTTP(s) server.
+	if s.opts.NATSServerURL != "" {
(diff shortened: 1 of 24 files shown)


server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+	"strconv"
+	"time"
+
+	gnatsd "github.com/nats-io/gnatsd/server"
+	"github.com/nats-io/nats-streaming-server/stores"
+)
+
+// Routes for the monitoring pages
+const (
+	RootPath     = "/nss"
```
petemiron: should this be a full `streaming` instead of nss? we just don't use nss anywhere else.

server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+	ServerPath   = RootPath + "/server"
```
petemiron: should these names follow the same convention(z) as gnatsd (`varz`, `storez`, `clientz`, `channelz`)?

server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+	mux.HandleFunc(RootPath, s.handleRootz)
+	mux.HandleFunc(ServerPath, s.handleServerz)
+	mux.HandleFunc(StorePath, s.handleStorez)
+	mux.HandleFunc(ClientsPath, s.handleClientsz)
+	mux.HandleFunc(ChannelsPath, s.handleChannelsz)
+
+	return nil
+}
+
+func (s *StanServer) handleRootz(w http.ResponseWriter, r *http.Request) {
```
petemiron: Can we hijack existing root route (`/`) and append either a link to all of the additional routes, or append a link to `nss` under a heading or link to `streaming`?

server/monitor.go
kozlovic: sure

server/monitor.go
kozlovic: I did that at the beginning but then removed it. I will put it back.

server/monitor.go
kozlovic: No we can't. The `HandleFunc` would return the error `http: multiple registrations for ...`, this is why I had to have a "sub" root for streaming.

So the idea is that the doc (and banner on startup?) would refer to "http://localhost:port/streaming". One can then manually go to NATS monitoring by removing `streaming`. Note that there is situations where NATS monitoring would report nothing when the Streaming server does not embed the NATS server.

petemiron:
What if you did a reflect on all the registered handlers and just made this
a dynamic index page?

server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+		}
+		sendResponse(w, r, clientsz)
+	}
+}
+
+func getMonitorClient(c *stores.Client, subsOption int) *Clientz {
+	cli := c.UserData.(*client)
+	cli.RLock()
+	if cli.unregistered {
+		cli.RUnlock()
```
tylertreat: Seems like a single `defer cli.RUnlock()` would be appropriate in this function since it's not really a hotpath.

server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+func (a byChannelName) Less(i, j int) bool { return a[i].Name < a[j].Name }
+
+func (s *StanServer) handleChannelsz(w http.ResponseWriter, r *http.Request) {
+	oneChannel := r.URL.Query().Get("channel")
+	subsOption, _ := strconv.Atoi(r.URL.Query().Get("subs"))
+	if oneChannel != "" {
+		cs := s.store.LookupChannel(oneChannel)
+		if cs == nil {
+			w.WriteHeader(http.StatusNotFound)
+			w.Write([]byte(fmt.Sprintf("Channel %s not found", oneChannel)))
```
tylertreat: https://golang.org/pkg/net/http/#Error

server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+func (a byName) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a byName) Less(i, j int) bool { return a[i] < a[j] }
+
+type byChannelName []*Channelz
+
+func (a byChannelName) Len() int           { return (len(a)) }
+func (a byChannelName) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a byChannelName) Less(i, j int) bool { return a[i].Name < a[j].Name }
+
+func (s *StanServer) handleChannelsz(w http.ResponseWriter, r *http.Request) {
```
tylertreat: This function is fairly large with lots of branching and could use some decomposing for readability/maintainability purposes IMO.

server/monitor.go
kozlovic: No idea.. comes from NATS Server, that's all I can say :-)

server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+	PendingCount int    `json:"pending_count"`
+	IsStalled    bool   `json:"is_stalled"`
+}
+
+func (s *StanServer) startMonitoring(nOpts *gnatsd.Options) error {
+	var hh http.Handler
+	// If we are connecting to remote NATS Server, we start our own
+	// HTTP(s) server.
+	if s.opts.NATSServerURL != "" {
+		s.natsServer = gnatsd.New(nOpts)
```
kozlovic: When not embedding the NATS server, we would have to duplicate the code starting the http server, but also the checks about the http ports (a PR that was recently merged in NATS Server), which now happen in gnatsd.StartMonitoring(). It is not *that* much code, but if we can avoid duplicating, I think there is merit. I am not so concern with the heavyweight of creating a new instance of NATS Server (the server is not started, not client accept loop, etc..), but I agree that the NATS Server specific endpoints would then always return empty values.

server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+		timeout := time.Now().Add(10 * time.Second)
+		for time.Now().Before(timeout) {
+			if hh = s.natsServer.HTTPHandler(); hh != nil {
+				break
+			}
+			time.Sleep(250 * time.Millisecond)
+		}
+	}
+	if hh == nil {
+		return fmt.Errorf("unable to start monitoring server")
```
kozlovic: I confess that I now always use `fmt.Errorf()`, even without formatting, because in most case I already import `fmt`. But agreed, from now on I will try to use `errors.New()` when there is no formatting required.

server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+	subsOption, _ := strconv.Atoi(r.URL.Query().Get("subs"))
+	if singleClient != "" {
+		var clientz *Clientz
+		client := s.store.GetClient(singleClient)
+		if client != nil {
+			clientz = getMonitorClient(client, subsOption)
+		}
+		if clientz == nil {
+			w.WriteHeader(http.StatusNotFound)
+			w.Write([]byte(fmt.Sprintf("Client %s not found", singleClient)))
```
kozlovic: ok, thank.

server/monitor.go
kozlovic: done

server/monitor.go
kozlovic: With the change of having a single Channelsz, it may be shorter, but I see your point.

server/monitor.go
kozlovic: Sure, there should not be million of clients (because this is call for every client in the list of clients returned by the `/clientsz` endpoint).

As a general rule, though, I don't use defer if it is fast path, and if the code complexity is so low that it is easy to not miss the unlock. For instance, for a function that would do something as simple as returning an object's field, I would prefer this to a defer:
```
func (s *mystruct) getObj() *obj {
  s.Lock()
  o := s.obj
  s.Unlock()
  return o
}
```  
(I may have written some code that does just the opposite, but the idea is that it's not because the language provides defer that you have to use it all the times. And again, if that function was called only once in a while, it does not matter, but too often, and I would really prefer not to use the defer).

server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+type ChannelsWithSubsz struct {
+	ClusterID string      `json:"cluster_id"`
+	ServerID  string      `json:"server_id"`
+	Now       time.Time   `json:"now"`
+	Offset    int         `json:"offset"`
+	Limit     int         `json:"limit"`
+	Count     int         `json:"count"`
+	Total     int         `json:"total"`
+	Channels  []*Channelz `json:"channels"`
+}
```
kozlovic: done

kozlovic:
@petemiron @tylertreat I cannot override an handler already set, and I don't think (did not find how to) programmatically find out what are all registered handlers. Anyway, the later would not help unless I could override the root handler (`/`) so that I could display a new generated page that would contain all NATS + Streaming endpoints.

Other possible approaches:
1. Streaming we could register an "index.html" that will list all NATS and Streaming endpoints in one page, but that would be a departure from NATS (in that we don't use one for NATS Server). I also would need to update manually this page if new endpoints are added to NATS server - which is ok since we vendor it, because again, I am not sure how we can programmatically find out which one are already registered. When not embedding the NATS Server - but using its http server as it is done right now - we could also omit, in that landing page, the NATS specific endpoints (although the embedded NATS Server would still service them) since they are meaningless in such case.
2. Use http/https/host options just for Streaming, and they would have to be different from NATS server. Which means that if one embeds NATS Server and still want to monitor for both Streaming and NATS, one would need to provide 2 ports. Each URL having different ports, they could both have `/` for the root, but you would not be able to get every endpoints links from a central page.

server/monitor.go
```diff
@@ -0,0 +1,504 @@
...
+
+		minoff := offset
+		if minoff > totalClients {
+			minoff = totalClients
+		}
+		maxoff := offset + limit
+		if maxoff > totalClients {
+			maxoff = totalClients
+		}
+		carr = carr[minoff:maxoff]
```
kozlovic: But at the end, since both store implementations maintain channels (same for clients) as a map, in order to return a subset, we would still need to create an array with all channels in order to sort and then apply the offset/limit and return a subset of this array.
Even with adding an internal number on a channel structure, we would need to store the channels in two maps, one per name, one per number and that could be used as the offset key. Problem is for clients (that can be removed) and will be for channels when we add the ability to remove them.
For channels, after a server restart, the numbering would not be the same since recovery will not guarantee the same ordering than when channels were first created. So it means that getting offset/limit between server restart would return inconsistent value. With current ordering based on name, it will still be ok.
What do you think?

kozlovic:
@petemiron @tylertreat I have addressed most of the simple comments. There are 2 main points left:
1. is as I describe in previous comment: what to do with the handlers/landing page, etc..
2. do we move pagination to store for channels/clients, if so, do you have an idea on how to make this more efficient that it is right now.

PR #331, opened 2017-06-28
[UPDATED] Logging based on changes to gnatsd logger

kozlovic:
Logging in NATS Streaming has been reworked.


logger/logger.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (s *StanLogger) executeLogCall(f func(logger Logger, format string, v ...interface{}), format string, args ...interface{}) {
+	s.mu.Lock()
+	if s.log == nil {
+		s.mu.Unlock()
+		return
+	}
+	f(s.log, LogPrefix+format, args...)
+	s.mu.Unlock()
+}
```
tylertreat: This code looks pretty similar to the gnatsd logger. I think at some point we should think about making a NATS logger library.

server/server.go
```diff
@@ -688,30 +703,26 @@ var DefaultNatsServerOptions = server.Options{
...
-	Noticef("connection %q reconnected to NATS Server at %q",
+func (s *StanServer) stanReconnectedHandler(nc *nats.Conn) {
+	s.log.Noticef("connection %q reconnected to NATS Server at %q",
 		nc.Opts.Name, nc.ConnectedUrl())
 }
 
-func stanClosedHandler(nc *nats.Conn) {
-	Debugf("connection %q has been closed", nc.Opts.Name)
+func (s *StanServer) stanClosedHandler(nc *nats.Conn) {
+	if s.debug {
```
tylertreat: Why the check? Logger should handle the conditional logging, right?

server/server.go
```diff
@@ -821,7 +832,9 @@ func (s *StanServer) createNatsClientConn(name string, sOpts *Options, nOpts *se
 		ncOpts.ReconnectBufSize = 128
 	}
 
-	Tracef(" NATS conn opts: %v", ncOpts)
+	if s.trace {
```
tylertreat: Same question as above.

kozlovic: For performance. Early on, I realized that the cost in calling Tracef or Debugf is substantial in term of string allocation (if there is formatting involved, etc..), so it was better to do the check at the higher layer. I have kept the check at the lower level in case one forgets to do the check before calling those functions, and to respect the contract of the logger.

logger/logger_test.go
```diff
@@ -0,0 +1,117 @@
+// Copyright 2016 Apcera Inc. All rights reserved.
```
kozlovic: Will fix the copyright year here

ColinSullivan1:
@kozlovic , was there an issue, or is this maintenance?

server/server.go
tylertreat: Should we make note of this in a comment? It might not be obvious to future maintainers.

stores/limits.go
```diff
@@ -19,16 +19,23 @@ const (
...
-		for k, v := range sl.PerChannel {
-			copyVal := *v
-			cloned.PerChannel[k] = &copyVal
-		}
-	}
+	cloned.PerChannel = sl.ClonePerChannelMap()
 	return &cloned
 }
 
+// ClonePerChannelMap returns a deep copy of the StoreLimits's PerChannel map
```
ColinSullivan1: IMO this should be in a different PR; it's not logging related.

server/server.go
kozlovic: But there are many places in the code where I do that. It happens that you saw this one because of the way the code review is.

kozlovic: If you want to check what the server's `debug` and `trace` flags are, you would find this in the `StanServer` structure. I think that's enough:
```
	// Use these flags for Debug/Trace in places where speed matters.
	// Normally, Debugf and Tracef will check an atomic variable to
	// figure out if the statement should be logged, however, the
	// cost of calling Debugf/Tracef is still significant since there
	// may be memory allocations to format the string passed to these
	// calls. So in those situations, use these flags to surround the
	// calls to Debugf/Tracef.
```
I will remove the `atomic` reference. In this PR, I also surrounded the Debug print statements about subjects with `if s.debug`, but that was unnecessary. As the comment suggest, you would want to do that only when speed matters (that is, in the fast path). I am going to look at the place where I use the server's debug/trace and remove it in place where it is not really needed.

kozlovic:
@tylertreat @ColinSullivan1 I have removed the checks to s.debug/trace where not needed and removed the Clone() usage for this PR (once this is merged, I will submit a new one).
Let me know if this is ok for merge. Thanks!

tylertreat:
lgtm

PR #332, opened 2017-06-29
Added Options.Clone()

kozlovic:
In order for the server to properly own the given options,
implement a Clone() method.


kozlovic:
I am merging this one without explicit code review because it was reviewed in PR about logging but deemed not to belong in that one.

PR #339, opened 2017-07-11
[CHANGED] MsgStore APIs now return errors

kozlovic:
This would allow the server to report possible failures.


server/monitor.go
```diff
@@ -431,7 +431,7 @@ func (s *StanServer) handleOneChannel(w http.ResponseWriter, r *http.Request, na
 
 func updateChannelz(cz *Channelz, cs *stores.ChannelStore, subsOption int) {
 	msgs, bytes, _ := cs.Msgs.State()
-	fseq, lseq := cs.Msgs.FirstAndLastSequence()
+	fseq, lseq, _ := cs.Msgs.FirstAndLastSequence()
```
tylertreat: What is the reason for ignoring the error here?

server/server.go
```diff
@@ -529,8 +531,10 @@ func (ss *subStore) Remove(cs *stores.ChannelStore, sub *subState, unsubscribe b
 			storageUpdate = sub.LastSent == qs.lastSent
 			sortedPendingMsgs := makeSortedPendingMsgs(sub.acksPending)
 			for _, pm := range sortedPendingMsgs {
-				m := cs.Msgs.Lookup(pm.seq)
-				if m == nil {
+				m, err := cs.Msgs.Lookup(pm.seq)
+				if m == nil || err != nil {
+					ss.stan.log.Errorf("Unable to update subscription for %s:%v (%v)",
+						sub.subject, pm.seq, err)
```
tylertreat: Why log here rather than bubbling the error up?

server/server.go
```diff
@@ -529,8 +531,10 @@ func (ss *subStore) Remove(cs *stores.ChannelStore, sub *subState, unsubscribe b
 			storageUpdate = sub.LastSent == qs.lastSent
 			sortedPendingMsgs := makeSortedPendingMsgs(sub.acksPending)
 			for _, pm := range sortedPendingMsgs {
-				m := cs.Msgs.Lookup(pm.seq)
-				if m == nil {
+				m, err := cs.Msgs.Lookup(pm.seq)
+				if m == nil || err != nil {
```
tylertreat: In what situations is `m` nil and `err` not nil?

server/server.go
```diff
@@ -3319,17 +3329,21 @@ func (s *StanServer) sendAvailableMessages(cs *stores.ChannelStore, sub *subStat
...
+		}
 		if nextMsg != nil {
 			return nextMsg
 		}
 		// Reason why we don't call FirstMsg here is that
 		// FirstMsg could be costly (read from disk, etc)
 		// to realize that the message is of lower sequence.
 		// So check with cheaper FirstSequence() first.
-		firstAvail := cs.Msgs.FirstSequence()
+		firstAvail, _ := cs.Msgs.FirstSequence()
```
tylertreat: Why ignore the error?

server/server.go
```diff
@@ -3349,13 +3363,19 @@ func getNextMsg(cs *stores.ChannelStore, nextSeq, lastSent *uint64) *pb.MsgProto
 	}
 }
 
-func (s *StanServer) getSequenceFromStartTime(cs *stores.ChannelStore, startTime int64) uint64 {
-	return cs.Msgs.GetSequenceFromTimestamp(startTime)
+func (s *StanServer) getSequenceFromStartTime(cs *stores.ChannelStore, startTime int64) (uint64, error) {
+	tm, err := cs.Msgs.GetSequenceFromTimestamp(startTime)
+	if err != nil {
+		s.log.Errorf("Unable to get sequence from start time: %v", err)
+		return 0, err
```
tylertreat: Is this potential for double logging since we are logging here *and* propagating the error where it might be logged higher in the stack?

server/server.go
kozlovic: We either need to panic/fatal, since we can't "move" this message to a different queue member, or log and continue. Same happens few lines below with the call to `AddSeqPending`.

server/server.go
kozlovic: That will be in another PR where I need to handle gaps. For now simply ignore.

server/server.go
kozlovic: I will, but the top layer may log something like "unable to add subscription: %v", and so this trace is adding more information: it happened when trying to get the sequence start. That message itself will contain the reason for the failure coming from the store. Unless we change from error to something that allows us to add causes (like a stack), we will always have this issue.

kozlovic: But ok, will remove. Actually, will remove that function. The caller of this one will call this directly and will bubble up the error to caller.

stores/common_test.go
```diff
@@ -393,14 +452,14 @@ func testMaxMsgs(t *testing.T, s Store) {
 	}
 
 	// Check that older messages are no longer avail.
-	if cs.Msgs.Lookup(1) != nil || cs.Msgs.Lookup(uint64(firstSeqAfterLimitReached-1)) != nil {
+	if msgStoreLookup(t, cs.Msgs, 1) != nil || msgStoreLookup(t, cs.Msgs, uint64(firstSeqAfterLimitReached-1)) != nil {
```
petemiron: suggest using a new line for each Or to assist in readability:
```
if msgStoreLookup(t, cs.Msgs, 1) != nil || 
     msgStoreLookup(t, cs.Msgs, uint64(firstSeqAfterLimitReached-1)) != nil {
```

stores/common_test.go
```diff
@@ -758,10 +817,13 @@ func TestGSNoOps(t *testing.T) {
 	gms := &genericMsgStore{}
 	defer gms.Close()
 	gms.init("foo", testLogger, &limits.MsgStoreLimits)
-	if gms.Lookup(1) != nil || gms.FirstMsg() != nil || gms.LastMsg() != nil || gms.Flush() != nil ||
-		gms.GetSequenceFromTimestamp(0) != 0 || gms.Close() != nil {
+	if msgStoreLookup(t, gms, 1) != nil || msgStoreFirstMsg(t, gms) != nil || msgStoreLastMsg(t, gms) != nil ||
```
petemiron: suggest multiline Or.

stores/common_test.go
```diff
@@ -758,10 +817,13 @@ func TestGSNoOps(t *testing.T) {
...
+		gms.Flush() != nil || msgStoreGetSequenceFromTimestamp(t, gms, 0) != 0 || gms.Close() != nil {
```
petemiron: suggest multiline Or.

PR #345, opened 2017-07-25
[FIXED] FileStore's GetSequenceFromTimestamp panic

### LINKED ISSUE #344: Instance panicks periodically
Hi folks, 

My production instance of nats-streaming panicks every few day.
I'm still investigating this but I'm filing this first just in case you folks know what happened.

The deployment environment is as follow.
The said problematic instance is deployed as a Stateful Set at Kubernetes 1.5, 
using docker image nats-streaming:0.5.0.
The file storage is mounted at Kubernetes NFS Volume.

It might be a few factors, I suspect it is either NFS or when I gzipped my payload (by correlation, it started after I did that). I suspected file sync too, but it still happens after I turned it on.

Here's the log (for the recovery attempts), it happens randomly while running too.
https://pastebin.com/SCW2XQwh

kozlovic:
This would happen if all messages in a channel have expired.

Resolves #344

### DIFF
--- a/stores/common_test.go
+++ b/stores/common_test.go
@@ -686,6 +686,9 @@ func testBasicSubStore(t *testing.T, s Store) {
 }
 
 func testGetSeqFromStartTime(t *testing.T, s Store) {
+	limits := testDefaultStoreLimits
+	limits.MaxAge = 500 * time.Millisecond
+	s.SetLimits(&limits)
 	// Force creation of channel without storing anything yet
 	s.CreateChannel("foo", nil)
 	// Lookup channel store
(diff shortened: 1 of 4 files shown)


stores/common_test.go
```diff
@@ -686,6 +686,9 @@ func testBasicSubStore(t *testing.T, s Store) {
 }
 
 func testGetSeqFromStartTime(t *testing.T, s Store) {
+	limits := testDefaultStoreLimits
+	limits.MaxAge = 500 * time.Millisecond
```
petemiron: in the sake of test timing, any reason not to set this *really* low? like 10millis?

kozlovic: If you look down, you will see that we are sending 100 messages, each one with a 1ms delay. So no, 10ms would not work. Also, be careful to be too tight because then you start to get Travis flappers ;-)

PR #346, opened 2017-07-26
[CHANGED] Store interface refactoring

kozlovic:
Some of channels and clients responsibilities have been moved
to the server, leading to removal of some of the Store APIs.


README.md
```diff
@@ -56,12 +56,31 @@ NATS Streaming provides the following high-level feature set.
 
 ## Version `0.6.0`
 
-The SubStore interface was updated.
+The Store interface has been heavily modified. Some of the responsibilities have been moved into the server
```
petemiron: We should clarify if files are backwards compatible.

PR #348, opened 2017-07-29
[CHANGED] Debug/trace statements related to Subscriptions

kozlovic:
Improve the debug statement when creating/destroying a subscription.
Report the sub id and use it in all other subscription related
debug/trace statement to better correlate actions from this
subscription (the inbox was too chatty and difficult to read).

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -544,6 +544,8 @@ func (ss *subStore) Remove(c *channel, sub *subState, unsubscribe bool) {
 	sub.Lock()
 	subject := sub.subject
 	inbox := sub.Inbox
+	clientID := sub.ClientID
+	durName := sub.DurableName
 	sub.removed = true
 	sub.clearAckTimer()
 	durableKey := ""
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -3246,6 +3293,45 @@ func (s *StanServer) processSubscriptionRequest(m *nats.Msg) {
 	s.subStartCh <- &subStartInfo{c: c, sub: sub, qs: qs, isDurable: isDurable}
 }
 
+func traceSubscriptionRequest(log logger.Logger, sr *pb.SubscriptionRequest,
```
petemiron: Probably useful to add a unit test for this debug trace string. Could be added into existing tests.

server/server.go
```diff
@@ -720,6 +730,46 @@ func (ss *subStore) Remove(c *channel, sub *subState, unsubscribe bool) {
...
 		ss.stan.performAckExpirationRedelivery(qsub)
 	}
+
+	if log != nil {
+		traceCloseOrUnsubscribeRequest(log, clientID, subject, inbox, qgroup, durName, subid,
+			queueGroupIsEmpty, isDurable, unsubscribe)
+	}
+}
+
+func traceCloseOrUnsubscribeRequest(log logger.Logger, clientID, subject, inbox, qgroup, durName string,
```
petemiron: consider adding unit tests for new string logic.

server/server.go
petemiron: Parallel logic with traceUnsubscribeOrCloseRequest, possible to combine funcs?

PR #349, opened 2017-07-31
Use GetDefaultOptions() instead of DefaultOptions from go-nats.

nogoegst:


nogoegst:
This should also include vendored changes from https://github.com/nats-io/go-nats-streaming/pull/145.

kozlovic:
@nogoegst NATS Streaming Server vendors the NATS (server and) go-client. Will have to update only after the vendors are updated. Closing this for now.

PR #376, opened 2017-08-15
[FIXED] Stalled Queue member may stall the whole Queue

### LINKED ISSUE #375: Stalled Queue member may stall the group
If a queue member is stalled, it may still be picked as the member to
send a message to, even when there are other members that are not
stalled, effectively stalling the whole group.

kozlovic:
This happens if a member is stalled and other members are not but
have a number of outstanding acks higher than the stalled member.

Resolves #375

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2181,7 +2181,9 @@ func (s *StanServer) sendPublishErr(subj, guid string, err error) {
 
 // FIXME(dlc) - place holder to pick sub that has least outstanding, should just sort,
 // or use insertion sort, etc.
-func findBestQueueSub(sl []*subState) (rsub *subState) {
+func findBestQueueSub(sl []*subState) *subState {
+	var rsub *subState
+
 	for _, sub := range sl {
 
 		if rsub == nil {
@@ -2203,7 +2205,7 @@ func findBestQueueSub(sl []*subState) (rsub *subState) {
 
 		// Favor non stalled subscribers and clients that do not have
 		// failed heartbeats
-		if (!sStalled || rStalled) && (!sHasFailedHB || rHasFailedHB) && (sOut < rOut) {
+		if (!sStalled && !sHasFailedHB && (rStalled || rHasFailedHB)) || (sOut < rOut) {
 			rsub = sub
 		}
 	}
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -2187,7 +2189,7 @@ func findBestQueueSub(sl []*subState) (rsub *subState) {
 
 		// Favor non stalled subscribers and clients that do not have
 		// failed heartbeats
-		if (!sStalled || rStalled) && (!sHasFailedHB || rHasFailedHB) && (sOut < rOut) {
+		if (!sStalled && !sHasFailedHB && (rStalled || rHasFailedHB)) || (sOut < rOut) {
```
petemiron: This logic takes a bit to grok and got me thinking about the func as a whole. Could this be simplified to make the criteria evaluation for "BestQueueSub" more clear and reduce locking:

```
// ... (line 2170)
leastOutstanding := int(^uint(0) >> 1) // set to max int
for _, sub := range sl {

	sub.RLock()
	sOut := len(sub.acksPending)
	sStalled := sub.stalled
	sHasFailedHB := sub.hasFailedHB
	sub.RUnlock()

	if !sStalled && !sHasFailedHB {
		if sOut < leastOutstanding {
			leastOutstanding = sOut			
			rsub = sub
		}
	}
}
...
```

I realize this would increase the time between evaluating a subscriber's state and possibly sending a message to it, but it would also reduce the locks so hopefully the loop would run a tiny bit faster.

kozlovic: Yes. The only difference in behavior is that with the current one, we at least pick one (even if it is stalled) as long as the array has 1 element. In the above approach (as-is), we would potentially return `nil`. (just so you know, I am working on a change for better handling `stalled` semantic, especially for queue groups, which then could address the fact that we return `nil` here if all are stalled).

kozlovic: But for this PR and to maintain current behavior, we may want to pick the first at the end of the loop if `rsub` has not been set in the loop.

petemiron: if you're going to change the behavior shortly, I wouldn't recommend this change. I'll approve.

kozlovic: I have to return a sub, because this function is also used in case of redeliveries where redeliveries are always forced (even if the sub is stalled). If I don't return a sub, test `TestPersistentStoreStalledQueueRedelivery` would fail.

PR #378, opened 2017-08-17
[IMPROVED] Stalled (queue) subscriptions

kozlovic:
- When new message arrives in channel and sent to subscription,
  a lookup was done even if the subscription was stalled. This
  is no longer the case.
- A queue subscription is stalled only when all its members are
  stalled. Again, no lookup necessary if the group is stalled.
- Removal of "stalled" debug statement that would be printed
  for every message after a subscription has reached its
  MaxInflight.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -412,11 +412,11 @@ type subStore struct {
 // tracks lastSent for the group.
 type queueState struct {
 	sync.RWMutex
-	lastSent  uint64
-	subs      []*subState
-	shadow    *subState // For durable case, when last member leaves and group is not closed.
-	stalled   int       // number of stalled members
-	newOnHold bool
+	lastSent        uint64
+	subs            []*subState
+	shadow          *subState // For durable case, when last member leaves and group is not closed.
+	stalledSubCount int       // number of stalled members
+	newOnHold       bool
 }
 
 // When doing message redelivery due to ack expiration, the function
@@ -538,7 +538,7 @@ func (ss *subStore) updateState(sub *subState) {
 		}
 		// Update stalled (on recovery)
 		if sub.stalled {
-			qs.stalled++
+			qs.stalledSubCount++
 		}
 		qs.Unlock()
 		sub.qstate = qs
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -415,7 +415,7 @@ type queueState struct {
 	lastSent  uint64
 	subs      []*subState
 	shadow    *subState // For durable case, when last member leaves and group is not closed.
-	stalled   bool
+	stalled   int       // number of stalled members
```
petemiron: recommend making this variable name more descriptive `stalledMemberCnt` or `stalledSubCnt`. parsing some of the latter logic of subscriber stalled flag vs. this stalled var becomes hard to follow.

server/server.go
```diff
@@ -3430,10 +3427,14 @@ func (s *StanServer) processAck(c *channel, sub *subState, sequence uint64) {
 		return
 	}
 
-	delete(sub.acksPending, sequence)
-	stalled := sub.stalled
-	if int32(len(sub.acksPending)) < sub.MaxInFlight {
-		sub.stalled = false
+	// Do this here only if this is not a queue member. If it is, it has
```
petemiron: consider setting the `qs` variable before this `if`, then converting this to an `if/else` instead of 2 separate `if`s.

server/server.go
```diff
@@ -3443,8 +3444,21 @@ func (s *StanServer) processAck(c *channel, sub *subState, sequence uint64) {
 
 	if qs != nil {
 		qs.Lock()
-		stalled = qs.stalled
-		qs.stalled = false
+		stalled = qs.stalled == len(qs.subs)
```
petemiron: specific example of hard-to-parse logic.

server/server.go
kozlovic: Note that the sub's lock need to be released before grabbing the queue state's lock. Still, I will update based on your comment and you will decide if it is better or not.

server/server.go
kozlovic: I did that originally, so I agree. I'll go for `stalledSubCount`.

server/server.go
kozlovic: agreed

PR #383, opened 2017-08-29
Clustering MVP [WIP]

tylertreat:
Sorry for the big PR. Note that this is going into an integration branch for the clustering MVP to allow for incremental additions, reviews, feedback, etc.

This initial PR does a few things:
- Add a Raft group per channel
- Channel leader is responsible for sequencing and replicating messages to followers via Raft protocol
- Channel leader now assigns message sequence numbers before replicating (note that Raft will prevent gaps)
- Channel leader is responsible for handling subscription requests (subscription "failover" and ack replication are not yet supported)
- Make store implementations idempotent
- Naive snapshotting of channels for Raft catch-up
- Add a basic clustering test which verifies replication works as expected even in the face of failures

@kozlovic @petemiron @ColinSullivan1


server/server.go
```diff
@@ -236,15 +256,157 @@ func (cs *channelStore) createChannel(s *StanServer, name string) (*channel, err
...
+		select {
+		case isLeader := <-node.LeaderCh():
+			if isLeader {
+				// Use a barrier to ensure all preceding operations are
+				// applied to the FSM, then update nextSequence.
+				if err := node.Barrier(0).Error(); err != nil {
+					// TODO: probably step down as leader?
+					panic(err)
+				}
+				lastSequence, err := c.store.Msgs.LastSequence()
```
kozlovic: If I understand correctly, we are here if the node was a follower and now becomes leader.
It is setting its nextSequence based on what is the last sequence from the store.
We use atomic for nextSequence because in the ioLoop, we may be in a 'replicate' call
that uses this field. What if here we were posting a control message which is processed 
the clientPublish handler. We could schedule a special iopm message to the ioLoop.
when processing this, we notify a channel the code will be waiting on here.
This would allow us not to use atomics for nextSequence.
I do something similar in remove_channel branch since I need to ensure that all
pending messages need to be done with a given channel before processing the request.
I could experiment here and will let you know if it is worth it.

server/server.go
```diff
@@ -3090,13 +3484,23 @@ func (s *StanServer) processSubscriptionRequest(m *nats.Msg) {
 		}
 	}
 
+	// In clustered mode, drop the request if we are not the channel leader.
```
kozlovic: I guess this was meant to be done below

stores/store.go
```diff
@@ -230,7 +230,7 @@ type MsgStore interface {
 	State() (numMessages int, byteSize uint64, err error)
 
 	// Store stores a message and returns the message sequence.
-	Store(data []byte) (uint64, error)
+	Store(msg *pb.MsgProto) (uint64, error)
```
kozlovic: Are you passing a `pb.MsgProto` because you need, or because it would break filestore otherwise?

If we are to change the Store() here, I wonder if we should not change it to `seq, []byte`, and Store
would store a bag of bytes with given sequence.

On Lookup() it would return bag of bytes for this sequence (again, filestore would need fixing and 
probably not be backward compatible).

Basically, it would remove MsgProto knowledge to the Store implementation.
The server would call Store() with the serialized version of MsgProto. This would allow, on Lookup,
to send the byte array directly to subs without the need to marshal it. We would have to do unmarshal/
marshal on redelivery since we need to set the Redelivered flag in the MsgProto though.

stores/common_test.go
```diff
@@ -175,9 +175,14 @@ func storeDeleteClient(t tLogger, s Store, clientID string) {
 	}
 }
 
-func storeMsg(t *testing.T, cs *Channel, channel string, data []byte) *pb.MsgProto {
+func storeMsg(t *testing.T, cs *Channel, channel string, seq uint64, data []byte) *pb.MsgProto {
```
kozlovic: I wonder if you could not have used a global variable for the sequence that is auto incremented.
For tests that expect sequences to start at a given number, then the test could set the global 
sequence to whatever value at beginning of function.
This is just an idea to reduce the number of changes in the test suite. Now that it is done, it is
kind of a moot point.

server/clustering_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+	sub.Unsubscribe()
+
+	// Take down the leader.
+	leader.Shutdown()
+	stopped = append(stopped, leader)
+	servers = removeServer(servers, leader)
+
+	// New subscriptions should be rejected since no leader can be established.
+	_, err = sc.Subscribe(channel, func(msg *stan.Msg) {}, stan.DeliverAllAvailable())
```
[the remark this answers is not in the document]
tylertreat: This is the downside of relying on interest-based pub/sub for NATS Streaming. This is *supposed* to timeout because there is no leader, but I'm not sure how we can fast-path this.

server/clustering_test.go
```diff
@@ -0,0 +1,323 @@
...
+	if err == nil {
+		t.Fatal("Expected error on subscribe")
+	}
+
+	// Publishes should also fail.
+	if err := sc.Publish(channel, []byte("foo")); err == nil {
```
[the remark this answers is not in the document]
tylertreat: Again, not sure how we can make this faster.

server/clustering_test.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+		for i := first; i <= last; i++ {
+			msg, err := store.Lookup(i)
+			if err != nil {
+				t.Fatalf("Error getting message %d: %v", i, err)
+			}
+			assertMsg(t, *msg, expected[i-1].data, expected[i-1].sequence)
+		}
+	}
+}
```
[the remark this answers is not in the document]
tylertreat: Yes, the test is quite slow. It's also doing a lot, but it's also slow due to testing timeouts. I'm open to ideas on how we can make it faster, though I think ultimately as we add more comprehensive test cases, I suspect we will need a way to run a separate suite of integration tests.

server/server.go
tylertreat: Oops, mistake. This comment should be removed.

stores/store.go
tylertreat: The reason for this change is the leader needs to assign metadata for a message (sequence and timestamp) before replicating. I'm open to suggestions here.

I like the idea of avoiding unnecessary marshals/unmarshals where possible. My profiling showed that has a pretty major impact on performance.

stores/common_test.go
tylertreat: Issue is that several tests store messages for different channels, so a global variable would result in gaps, unless we maintained a global map containing sequences for each channel.

server/clustering_test.go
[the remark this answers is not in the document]
kozlovic: You can pass the `ConnectWait()` in `stan.Connect()` to reduce the timeout. The timeout is used on Connect and other internal Request/Reply protocols. The danger on reducing it too much is that it may work 100% of cases on local, but flap on Travis (or when running with `-race`)

server/clustering_test.go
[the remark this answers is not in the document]
kozlovic: You passed `PubAcWait(5*time.Second)` when creating the connection, this is why. Default is 30 seconds, but you could lower it to say 2 secs, but again, we need to make sure that it does not cause Travis to flap.

stores/common_test.go
kozlovic: understood. Again, don't bother since you already did the work of changing all tests.

stores/store.go
kozlovic: (some notes to help with making a decision)

If we were to change, I realize that we would have to pass sequence and timestamp (along with byte array) to the `Store()` call since the store would still need those 2 informations (for message expiration, and lookup), which in your branch it gets from the `MsgProto` passed as a parameter.

If `FileMsgStore` were to store the given serialized `MsgProto` but without knowledge that it is a `MsgProto`, newer server would not be able to recover datastore that were created with the `0.2.2 MVP` release (for which index files did not exist). Users would have to start it with say a `0.5.0` which would create the index files on recovery, before being able to use the newer version that would not be able to recover data files without index files. (since newer server would not know that content is a `MsgProto` from which it can get sequence and timestamp).
The `0.2.2 MVP` being so old, I don't think anybody is running this version, so that would be less of an issue.

`MemStore` would have to change too, since it has a map of `MsgProto` keyed by sequence. For expiration it uses the timestamp from the `MsgProto` values. So we would need to have a struct that contains timestamp and byte array. (it kinds of duplicate the storage of timestamp since in reality it would be in the serialized data `MemoryMsgStore` is storing, but..).

server/clustering_test.go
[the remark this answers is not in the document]
petemiron: agree on planning to separate unit and integration test suites.

server/clustering_test.go
```diff
@@ -0,0 +1,323 @@
...
+	// Configure third server.
+	s3sOpts := getTestDefaultOptsForClustering("c", []string{"a", "b"})
+	s3, err := RunServerWithOpts(s3sOpts, nil)
+	if err != nil {
+		t.Fatalf("Error starting server: %v", err)
+	}
+	defer s3.Shutdown()
+
+	servers := []*StanServer{s1, s2, s3}
+
```
kozlovic: I would iterate this array and call: `checkState(t, s, Clustered)`. This will check that the state is properly set.

server/server.go
```diff
@@ -290,9 +452,100 @@ func (cs *channelStore) count() int {
...
+	return atomic.LoadUint32(&c.leader) == 1
+}
+
+// Snapshot is used to support log compaction. This call should
+// return an FSMSnapshot which can be used to save a point-in-time
+// snapshot of the FSM. Apply and Snapshot are not called in multiple
+// threads, but Apply will be called concurrently with Persist. This means
+// the FSM should be implemented in a fashion that allows for concurrent
+// updates while a snapshot is happening.
+func (c *channel) Snapshot() (raft.FSMSnapshot, error) {
```
kozlovic: I do not see this function and Restore() being called. I assume this is invoked from raft. From code coverage it looks like these are not covered. Would it be possible to have a test that shows the effect of snapshot and restore?

tylertreat: Yes, this is called by Raft. We can probably (and should) test this. Will address...

tylertreat:
@kozlovic I addressed your feedback with the exception of testing Snapshot/Restore and the leadership issue you mentioned in Slack. Will work on those next.

server/server.go
```diff
@@ -290,9 +452,100 @@ func (cs *channelStore) count() int {
...
+func (c *channel) isLeader() bool {
+	return atomic.LoadUint32(&c.leader) == 1
+}
+
+// Snapshot is used to support log compaction. This call should
+// return an FSMSnapshot which can be used to save a point-in-time
+// snapshot of the FSM. Apply and Snapshot are not called in multiple
+// threads, but Apply will be called concurrently with Persist. This means
+// the FSM should be implemented in a fashion that allows for concurrent
+// updates while a snapshot is happening.
```
ColinSullivan1: Comments like these are much appreciated!

server/server.go
```diff
@@ -808,6 +1067,9 @@ type Options struct {
 	AckSubsPoolSize    int           // Number of internal subscriptions handling incoming ACKs (0 means one per client's subscription).
 	FTGroupName        string        // Name of the FT Group. A group can be 2 or more servers with a single active server and all sharing the same datastore.
 	Partitioning       bool          // Specify if server only accepts messages/subscriptions on channels defined in StoreLimits.
+	ClusterNodeID      string        // ID of the node within the cluster.
+	ClusterPeers       []string      // Cluster peer IDs.
```
ColinSullivan1: I notice there is a `raft.AddPeer(string)` and `raft.RemovePeer(string)` - that may come in useful for dynamic cluster changes someday.

server/server.go
```diff
@@ -2686,6 +3012,21 @@ func (s *StanServer) ioLoop(ready *sync.WaitGroup) {
 				remaining -= ioChanLen
 			}
 
+			replicateFutures := storeIOPendingMsgs(batch)
+
+			// If clustered, wait on the result of replication.
+			if s.isClustered() {
+				for i, future := range replicateFutures {
+					iopm = pendingMsgs[i]
+					if err := future.Error(); err != nil {
```
ColinSullivan1: Don't we need only a quorum to succeed, not all futures?  Or is that handled under the hood?

server/server.go
```diff
@@ -17,15 +22,19 @@ import (
...
 	"time"
 
+	"github.com/hashicorp/raft"
+	"github.com/hashicorp/raft-boltdb"
 	natsdLogger "github.com/nats-io/gnatsd/logger"
 	"github.com/nats-io/gnatsd/server"
 	"github.com/nats-io/go-nats"
 	"github.com/nats-io/go-nats-streaming/pb"
+	"github.com/nats-io/nuid"
+	"github.com/tylertreat/nats-on-a-log"
```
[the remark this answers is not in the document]
tylertreat: Yep

server/server.go
tylertreat: That's happening under the hood in Raft. Each of these futures represents a separate Raft operation (in this case, replication of a batch of messages within a channel).

tylertreat:
@ColinSullivan1 @kozlovic I believe I've addressed all feedback.

server/clustering_test.go
```diff
@@ -380,3 +392,65 @@ func TestClusteringNoPanicOnShutdown(t *testing.T) {
...
+	} else {
+		s1.Shutdown()
+		follower = s1
+	}
+
+	// Wait for leader to step down.
+	time.Sleep(time.Second)
+
+	// Ensure there is no leader now.
+	verifyNoLeader(t, channel, s1, s2)
```
kozlovic: Let's give verifyNoLeader a timeout so that we can suppress the time.Sleep() above. The helper function verifyNoLeader() will return as soon as it find no leader, or try again, up to the timeout. The retry interval should be rather small, like 100ms.

server/clustering_test.go
```diff
@@ -454,3 +456,119 @@ func TestClusteringLeaderFlap(t *testing.T) {
...
+		}
+	}
+
+	// Publish a message to force a timely catch up.
+	if err := sc.Publish(channel, []byte("11")); err != nil {
+		t.Fatalf("Unexpected error on publish: %v", err)
+	}
+
+	// Wait for the cluster to quiesce.
+	time.Sleep(6 * time.Second)
```
kozlovic: Instead of this arbitrary sleep time, could we make the loop below retry up to a timeout?
Goal is to minimize test time.

server/snapshot.go
```diff
@@ -0,0 +1,76 @@
...
+	for seq := first; seq <= last; seq++ {
+		msg, err = c.channel.store.Msgs.Lookup(seq)
+		if err != nil {
+			return
+		}
+		if msg == nil {
+			// Channel truncation has occurred while snapshotting.
+			err = errors.New("channel was truncated while snapshotting")
+			return
+		}
```
kozlovic: Improve tracing and code coverage with this code instead:
```
// If msg is nil, channel truncation has occurred while snapshotting.
if msg == nil {
    return errors.New("channel %q was truncated while snapshotting", c.channel.name)
}
```
Since you have named the returned error, any error that you return can be checked in the defer function.
In other words, say that you had named it `.. (retError error)`, then in your function you don't have to use `retErr =`, you can use local `err` if you want. If the function does return an error, you can check `if retError != nil` in the exit defer function.
That is, you could have:
```
defer func() {
   if retErr != nil {
      sink.Cancel()
   }
}()
...
if err := something(); err != nil {
   return err
}
```
This would cause the `if` statement in the `defer` function to evaluate to true.

server/snapshot.go
```diff
@@ -0,0 +1,76 @@
...
+		if err != nil {
+			return
+		}
+		_, err = sink.Write(data)
+		if err != nil {
+			return
+		}
+	}
+	err = sink.Close()
+	return
```
kozlovic: replace above with `return sink.Close()`. See comments above.

server/clustering_test.go
```diff
@@ -454,3 +456,119 @@ func TestClusteringLeaderFlap(t *testing.T) {
...
+		first, last, err := store.FirstAndLastSequence()
+		if err != nil {
+			t.Fatalf("Error getting sequence numbers: %v", err)
+		}
+		if first != 1 {
+			t.Fatalf("Unexpected first sequence: expected 1, got %d", first)
+		}
+		if last != 11 {
+			t.Fatalf("Unexpected last sequence: expected 11, got %d", last)
+		}
```
kozlovic: Getting failure locally and on Travis with:
```
=== RUN   TestClusteringLogSnapshotCatchup
--- FAIL: TestClusteringLogSnapshotCatchup (7.51s)
	clustering_test.go:563: Unexpected last sequence: expected 11, got 10
```

tylertreat: Seems that the timeout is not always sufficient. That said, it looks like Raft allows you to configure the commit timeout, which might alleviate this. https://godoc.org/github.com/hashicorp/raft#Config

server/clustering_test.go
```diff
@@ -0,0 +1,577 @@
...
+	s3 := runServerWithOpts(t, s3sOpts, nil)
+	defer s3.Shutdown()
+
+	servers := []*StanServer{s1, s2, s3}
+	for _, s := range servers {
+		checkState(t, s, Clustered)
+	}
+
+	// Create a client connection.
+	sc, err := stan.Connect(clusterName, clientName, stan.PubAckWait(2*time.Second))
```
kozlovic: This test often flaps on travis.
Remove the `PubAckWait()` for this connection since this publish is the one creating the channel, so it may take it. Leave it to the default. You are now using a second connection with smaller timeouts for failed cases, so we should be ok.

tylertreat: Done

tylertreat:
@kozlovic *Now* I think tests are passing finally.

PR #393, opened 2017-09-19
Fixed flapping tests

kozlovic:
These tests were flapping in clustering branch. Adding some timeout
for the check instead of fixed sleep before check. Increase the wait
before restarting server when checking redeliveries on startup.


tylertreat:
lgtm

PR #398, opened 2017-10-09
Clustering conn replication

tylertreat:
Implement replication for client connect/disconnect.

@kozlovic


server/server.go
```diff
@@ -2234,6 +2188,52 @@ func (s *StanServer) connectCB(m *nats.Msg) {
 	s.finishConnectRequest(client, req, m.Reply)
 }
 
+func (s *StanServer) processConnect(req *pb.ConnectRequest) (*client, bool, error) {
```
kozlovic: This is not enough in a follower. It needs to handle possible existing client and replace with new one if old is not replying, or we need to change the way the leader replicates. I have added a test to the PR that shows the problem.

tylertreat:
@kozlovic I pushed some changes that should address the client "refresh" case.

kozlovic:
@tylertreat Good start, however, as we discussed, the client health HB is set only on the leader - which is fine - but would need to be stopped when leadership is lost and started on leadership acquired. Also, when a client is closed due to lack of HBs, the leader need to replicate that state too.

tylertreat:
@kozlovic Fixed HB failovers and disconnect replication on expired HBs.

server/server.go
```diff
@@ -2418,8 +2411,26 @@ func (s *StanServer) processCloseRequest(m *nats.Msg) {
...
 
+	// In clustered mode, drop the request if we're not the metadata leader.
+	// Another server will handle it.
+	if s.isClustered() && !s.isMetadataLeader() {
+		return
+	}
+
+	// If clustered, thread operations through Raft.
+	if s.isClustered() {
+		if err := s.replicateConnClose(req.ClientID, m.Reply, true); err != nil {
```
kozlovic: Here, server is metaleader and for it - only - the m.Reply matters. That is, this is the client's reply inbox the connect response/error needs to be sent to. This should not be used by followers.

server/server.go
```diff
@@ -2418,8 +2411,26 @@ func (s *StanServer) processCloseRequest(m *nats.Msg) {
...
+		}
+	} else {
+		s.scheduleConnClose(req.ClientID, m.Reply)
+	}
+}
+
+func (s *StanServer) scheduleConnClose(clientID, reply string) {
 	// Create the control and corresponding NATS message
-	ctrlMsg, ctrlNatsMsg := s.createCtrlMsg(spb.CtrlMsg_ConnClose, true, m.Reply, []byte(req.ClientID))
+	ctrlMsg, ctrlNatsMsg := s.createCtrlMsg(spb.CtrlMsg_ConnClose, true, reply, []byte(clientID))
```
kozlovic: `reply` should not be set (or reset to empty) if the server processing this is not the leader.

server/server.go
```diff
@@ -2486,24 +2497,39 @@ func (s *StanServer) processCloseRequest(m *nats.Msg) {
...
 // performConnClose performs a connection close operation after all
 // client's pubMsg or client acks have been processed.
-func (s *StanServer) performConnClose(m *nats.Msg, clientID string) {
+func (s *StanServer) performConnClose(clientID, reply string) {
 	// The function or the caller is already locking, so do not use
 	// locking in that function.
 	if !s.closeClient(clientID) {
 		s.log.Errorf("Unknown client %q in close request", clientID)
-		s.sendCloseErr(m.Reply, ErrUnknownClient.Error())
+		s.sendCloseErr(reply, ErrUnknownClient.Error())
```
kozlovic: Do not send error back if it is not the metaleader.

server/server.go
```diff
@@ -2486,24 +2497,39 @@ func (s *StanServer) processCloseRequest(m *nats.Msg) {
...
 		return
 	}
 
 	resp := &pb.CloseResponse{}
```
kozlovic: Skip all that if not the metaleader.

tylertreat:
@kozlovic This only appeared to be the case for connection close. We were only sending response from leader on connect.

PR #401, opened 2017-10-11
Allow configuring fsync for Raft log

tylertreat:
Apply the fsync configuration from the file store to the Raft log
backend. This allows disabling disk syncing on every write to the
Raft log which improves performance. While this creates some risk,
the thought is if we are replicating to enough nodes, the replication
itself is sufficient for HA of data since the likelihood of more than a
quorum of nodes failing is low.

Note: we may want to consider disabling fsync as the default in
clustered mode for performance reasons, similar to what Kafka
does.

@kozlovic

### DIFF
--- a/server/conf.go
+++ b/server/conf.go
@@ -480,9 +480,6 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 		for i, peer := range sopts.Clustering.Peers {
 			sopts.Clustering.Peers[i] = strings.TrimSpace(peer)
 		}
-
-		// If clustered, override store sync configuration with cluster sync.
-		sopts.FileStoreOpts.DoSync = sopts.Clustering.Sync
 	}
 
 	if sopts.Clustering.RaftLogPath == "" {
--- a/server/conf.go
+++ b/server/conf.go
@@ -6,7 +6,6 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
-	"path/filepath"
 	"reflect"
 	"strconv"
 	"strings"
--- a/server/conf.go
+++ b/server/conf.go
@@ -437,6 +437,7 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 	fs.IntVar(&sopts.Clustering.LogCacheSize, "cluster_log_cache_size", DefaultLogCacheSize, "stan.Clustering.LogCacheSize")
 	fs.IntVar(&sopts.Clustering.LogSnapshots, "cluster_log_snapshots", DefaultLogSnapshots, "stan.Clustering.LogSnapshots")
 	fs.Int64Var(&sopts.Clustering.TrailingLogs, "cluster_trailing_logs", DefaultTrailingLogs, "stan.Clustering.TrailingLogs")
+	fs.BoolVar(&sopts.Clustering.Sync, "cluster_sync", false, "stan.Clustering.Sync")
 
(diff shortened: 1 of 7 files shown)


tylertreat:
@kozlovic I would think we would also want to default with filestore sync=false for clustering as well?

kozlovic:
@tylertreat Not sure about defaulting. What we could do is this: introduce a new server/clustering option Sync that would default to false. In clustering mode, when server creates store, it could override the FileStoreOptions.DoSync based on the server option. We would do same for any other type of store with similar behavior.
This means that the file store option "sync" would basically be ignored and instead we would use the server clustering options Sync value. Makes sense?

kozlovic:
And what I meant is that before creating the store (if running in clustering mode), the server would set `sOpts.FileStoreOpts.DoSync = sOpts.Clustering.Sync` (or whatever name we chose for that new option).

tylertreat:
That works for me.

tylertreat:
@kozlovic added a `cluster_sync` option and refactored clustering configuration a bit.

kozlovic:
@tylertreat Did we not say that you would override `sOpts.FileStoreOpts.DoSync` with this new value before creating the FileStore? I do not see that in this change.

kozlovic:
@tylertreat I see that you do that in ConfigureOptions, but wonder if it is not too early. I would probably do that before creating the store.

server/conf.go
```diff
@@ -475,14 +476,17 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp,
...
 	if len(stanClusterPeers) > 0 {
-		sopts.ClusterPeers = strings.Split(stanClusterPeers, ",")
-		for i, peer := range sopts.ClusterPeers {
-			sopts.ClusterPeers[i] = strings.TrimSpace(peer)
+		sopts.Clustering.Peers = strings.Split(stanClusterPeers, ",")
+		for i, peer := range sopts.Clustering.Peers {
+			sopts.Clustering.Peers[i] = strings.TrimSpace(peer)
 		}
+
+		// If clustered, override store sync configuration with cluster sync.
```
kozlovic: I would move this into server code, before creating the Store. There are users that may run server from code by providing options without invoking this config options.

kozlovic: It will probably apply to all processing you do here with cluster related options. Keep in mind that user *can* start the server from code using `RunServerWithOpts()`.

kozlovic:
@tylertreat Not sure if you saw my other comment about clustering options processing that you do in ConfigureOptions() such as the path, and peers. These may have to be done in server code (instead or too, up to you).

tylertreat:
@kozlovic which processing is that? The only processing is converting the peers string to a list.

tylertreat:
The peers processing is just converting the string to a list, which only seems relevant for the CLI. The default path logic could probably move into the server though.

server/server.go
```diff
@@ -1351,6 +1345,11 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 		nOpts = natsOpts.Clone()
 	}
 
+	// Default cluster Raft log path to ./<cluster-id>/<node-id> if not set.
+	if sOpts.Clustering.RaftLogPath == "" {
```
kozlovic: Move this down where you already test for isClustered() (to do the file store sync override), or move the file store override back up here. Point is, this should be done only in clustered mode.

kozlovic:
@tylertreat Fix the import issue in conf.go please.

tylertreat:
@kozlovic I think I've addressed all feedback on this.

PR #404, opened 2017-10-13
[FIXED] Server is not rejecting `/` in channel names

### LINKED ISSUE #403: Server should reject `/` in channel names
The `/` symbol in channel names causes issues for FileStore implementation.
The server should reject channel names containing such character for all
store implementations to be consistent.

Related to #402

kozlovic:
This was a problem with FileStore implementation since channels
are not meant to translate into sub directories (but be flat).

The server now rejects channels containing `/`, regardless of
the store type being used.

Resolves #403

### DIFF
--- a/server/conf_test.go
+++ b/server/conf_test.go
@@ -308,6 +308,7 @@ func TestParseWrongTypes(t *testing.T) {
 	expectFailureFor(t, "store_limits:{channels:{\"foo.*bar\":{}}}", wrongSubjErr)
 	expectFailureFor(t, "store_limits:{channels:{\"foo.>.>\":{}}}", wrongSubjErr)
 	expectFailureFor(t, "store_limits:{channels:{\"foo..bar\":{}}}", wrongSubjErr)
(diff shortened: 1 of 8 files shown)


server/conf_test.go
```diff
@@ -308,6 +308,7 @@ func TestParseWrongTypes(t *testing.T) {
 	expectFailureFor(t, "store_limits:{channels:{\"foo.*bar\":{}}}", wrongSubjErr)
 	expectFailureFor(t, "store_limits:{channels:{\"foo.>.>\":{}}}", wrongSubjErr)
 	expectFailureFor(t, "store_limits:{channels:{\"foo..bar\":{}}}", wrongSubjErr)
+	expectFailureFor(t, "store_limits:{channels:{\"foo/bar\":{}}}", wrongSubjErr)
```
petemiron: For consistency with other changes, should these `wrongSubjErr`s be changed to `wrongChanErr`?

PR #408, opened 2017-10-18
Clustering sub snapshot/restore

tylertreat:
Implement clustering snapshot/restore for subscriptions. Also fixed a bug in the verification of channel consistency in clustering unit tests.

@kozlovic


server/snapshot.go
```diff
@@ -22,56 +34,208 @@ func newChannelSnapshot(c *channel) raft.FSMSnapshot {
...
 		}
 	}
-	return sink.Close()
+	return nil
 }
 
-// Release is a no-op.
-func (c *channelSnapshot) Release() {}
+func (c *channelSnapshot) snapshotAcks(sink raft.SnapshotSink) error {
+	// TODO
```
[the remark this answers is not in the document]
tylertreat: Probably true.

tylertreat:
@kozlovic I will add a test to this. Then I will look at replicating acks (possibly as part of subscriptions) in a separate PR.

tylertreat:
@kozlovic On second thought, I'm not sure exactly how to instrument the test you describe. We can start a cluster, kill a follower, create a subscription and publish messages to it, force a Raft compaction on the leader (which triggers snapshot), then bring the follower back up. At this point, the leader will stream the snapshot to the follower to restore from, but how do we test the ack replication since the leader is still serving the subscription?

kozlovic:
If you have a follower that starts with no state, it should recover from snapshot, right? If so, then it will get the messages and subscriptions, but not the acks. So if the leader is then killed and the follower becomes leader, it should then try to redeliver all the messages it got from the subscription's snapshot since it does not have the ack for those.

tylertreat:
@kozlovic Correct, but we need a cluster of at least 3 nodes (otherwise no leader can be elected if leader fails). Thus, there's no way to guarantee the follower who recovered from the snapshot is elected leader.

We might be able to "hack" it as the Raft library allows a bootstrap mode where a single node can elect itself leader. Thus, we could create a cluster of 1 node, force a snapshot, and then add 2 nodes to the cluster who would restore from snapshot, then kill the leader. This way either node that gets elected will have recovered from a snapshot. I will see if this is feasible, unless you have a different idea.

kozlovic:
My idea was that start with cluster 3 nodes, then shutdown the 2 followers, erase their state and restart them, but the raft approach with single node should work too.
We need a test that checks for snapshot anyway, so if not now, we will need it later.

kozlovic:
But then, if we shutdown the 2 followers to clear their state and restart, the leader should no longer be leader and won't be able to reply to snapshot recovery requests, right?

tylertreat:
@kozlovic A leader election would happen, but the old leader should win if its log is ahead of the followers. However, I'm not sure how we can "clear" the state of the followers. Deleting their store would mean they wouldn't recover the channel and thus never join the Raft group.

kozlovic:
> However, I'm not sure how we can "clear" the state of the followers. Deleting their store would mean they wouldn't recover the channel and thus never join the Raft group

.. so it means you can't never add a new member to the group, or if a channel is created while a follower was not running it will never catch up?

tylertreat:
Yes, this is a problem we have today. If a node is offline, it has no way to discover that a channel was created and that it needs to join.

server/clustering_test.go
```diff
@@ -143,10 +143,10 @@ type msg struct {
 }
 
 func verifyChannelConsistency(t *testing.T, channel string, timeout time.Duration,
-	expectedFirstSeq, expectedLastSeq uint64, expectedMsgs []msg, servers ...*StanServer) {
+	expectedFirstSeq, expectedLastSeq uint64, expectedMsgs map[uint64]msg, servers ...*StanServer) {
 	deadline := time.Now().Add(timeout)
+OUTTER:
```
petemiron: sp: `OUTER`.

server/snapshot.go
```diff
@@ -22,56 +34,208 @@ func newChannelSnapshot(c *channel) raft.FSMSnapshot {
...
-			return fmt.Errorf("channel %q was truncated while snapshotting", c.channel.name)
+			batchPool.Put(batch)
+			return fmt.Errorf("channel %q was truncated while snapshotting", c.name)
 		}
-		data, err = msg.Marshal()
-		if err != nil {
-			return
+
+		// Previous batch is full, ship it.
+		if len(batch.Messages) == snapshotBatchSize {
```
petemiron: Would we ever want to add a timer in here? Is it possible that it could take longer to read/send a batch than a client timeout?

tylertreat: Can you clarify what you mean? When a snapshot is taken, it's streamed out to disk. After the snapshot is saved, the Raft log is truncated. Then later, if a follower needs to catch up, the leader streams the snapshot to the follower (if it doesn't have the actual log entries). The batching is just an optimization since we would otherwise spend a lot of time serializing/deserializing each individual message.

kozlovic:
@tylertreat Forget about the test since we are not snapshotting the pending acks, obviously the new leader would not have any pending message for the subscription, so there should not be redelivered messages. What we would test is that it starts delivering at the LastSent+1 sequence, which I believe we have a test for that.

PR #411, opened 2017-10-20
Include acksPending in subscription snapshot

tylertreat:
Adding acksPending to subscription snapshots.

@kozlovic Wondering if you have any ideas on testing this? The problem I see is we have no way to reliably failover leadership to the node that restored from the snapshot.


kozlovic:
I would cheat ;-) I would have setup with 3 servers, and shutdown one of the follower. Then I would do send/receive based on what you want to test. I would then restart the follower, which should then catchup from snapshot. Then I would shut them all down and restart the follower that catched-up in standalone mode. The file store should then be what I expect it to be (all that should have been replicated should be there).

server/snapshot.go
```diff
@@ -219,11 +215,18 @@ func (c *channel) restoreFromSnapshot(snapshot io.ReadCloser) error {
...
+				// when the follower becomes leader and attempts to
+				// redeliver messages.
+				acksPending[seq] = 0
+			}
 			sub := &subState{
 				SubState:    *fragment.Sub.State,
 				subject:     fragment.Sub.Subject,
 				ackWait:     computeAckWait(fragment.Sub.State.AckWaitInSecs),
-				acksPending: make(map[uint64]int64),
+				acksPending: acksPending,
```
kozlovic: This is good for runtime, but you are not persisting those acks on store. After call to addSubscription, should you persist those acks or not?

tylertreat: Yes, good catch. We need to persist them.

tylertreat:
@kozlovic I pushed a test. LMK if that's what you had in mind.

server/clustering_test.go
```diff
@@ -644,6 +644,159 @@ func TestClusteringLogSnapshotRestore(t *testing.T) {
...
+
+	// Kill a follower.
+	var follower *StanServer
+	for _, s := range servers {
+		if leader != s {
+			follower = s
+			break
+		}
+	}
+	follower.Shutdown()
```
kozlovic: I was thinking that the follower should be done before the events (pub/sub/acks) occur, otherwise you are not checking that the follower really recover from snapshot, it could have had the state from before being shutdown, no?

tylertreat: Fixed.

server/clustering_test.go
```diff
@@ -644,6 +644,161 @@ func TestClusteringLogSnapshotRestore(t *testing.T) {
...
+	follower.opts.Clustering.Peers = nil
+	follower = runServerWithOpts(t, follower.opts, nil)
+	defer follower.Shutdown()
+
+	// Verify the message is redelivered and, after acking, we receive the next
+	// message.
+	select {
+	case msg := <-ch:
+		msg.Ack()
+		assertMsg(t, msg.MsgProto, []byte("1"), 1)
```
kozlovic: What I don't understand is that you send message '2' to "force a timely catch up", so that message should be in `ch`. I don't see the place where you consume message '2' before the server restart, so this test should fail here...

tylertreat: Maxinflight is 1, and we don't ack, so it should not deliver 2.

PR #416, opened 2017-10-31
Implement snapshot/restore for client connections

tylertreat:
Implement Raft snapshot/restore for clients. This should be the remaining bulk of the code for clustering. Remaining work is mainly testing, improving configuration story, upgrading to the newly released HashiCorp Raft v1.0, and (potentially) switching to the fixed Raft group design.

*Implementation note:* I still haven't fully thought through potential implications with having the separate Raft groups for subscriptions and clients and if this could cause problems with snapshot/restore. Currently, the way this works is we simply snapshot client ID and HB inbox and then on restore, drop existing clients, restore from snapshot, and apply any preexisting subs to those.

@kozlovic


server/snapshot.go
```diff
@@ -17,13 +17,131 @@ import (
...
+	// (heartbeat inbox and client ID). Subscriptions are replicated through
+	// the channel Raft groups, so upon snapshot restore, the server will
+	// load any of the client's subscriptions from the store.
+	var buf [4]byte
+	for _, client := range s.clients.getClients() {
+		fragment := newFragment()
+		fragment.FragmentType = spb.RaftSnapshotFragment_Conn
+		clientSnap := clientSnapshotPool.Get().(*spb.ClientSnapshot)
+		client.RLock()
+		clientSnap.Info = &client.info.ClientInfo
```
kozlovic: I believe that the client's info is immutable so we should be ok here not making a copy of ClientInfo.

server/snapshot.go
```diff
@@ -17,13 +17,131 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	// Restore any subscriptions.
+	// QUESTION: are there potential race conditions due to how subscriptions
+	// and clients are handled on different Raft groups?
+	oldClient, ok := oldClients[client.info.ID]
+	if ok {
+		for _, sub := range oldClient.getSubsCopy() {
```
kozlovic: `getSubsCopy()` says: `At least Read-lock must be held by the caller.`.
So you should get the subs first like that:
```
  oldClient.RLock()
  subs := oldClient.getSubsCopy()
  oldClient.RUnlock()

  for _, sub := range subs {
  ...
```

kozlovic:
I missed adding the second comment in the review. The request for change is really about the getSubsCopy(), not getting the ClientInfo.

server/snapshot.go
tylertreat: Is this possible with protobuf? I think it requires a pointer.

server/snapshot.go
tylertreat: Good catch.

server/snapshot.go
kozlovic: Yes, since you use an embedded message, it is stored as pointer, what I meant is that you could do this:
```
copyInfo := client.info.ClientInfo
clientSnap.Info = &copyInfo
```
In that way, if client.info.ClientInfo struct content were to change, clientSnap.Info would not be affected. But again, I believe that this is immutable so we should be safe.

tylertreat: Ah yes, I misunderstood your original comment. I will leave this as is.

PR #422, opened 2017-11-03
Update raft and implement cluster bootstrapping

tylertreat:
This updates the raft dependency to the new 1.0.0 release and makes the necessary code changes to use that.

This also changes how clustering is configured. Instead of specifying your cluster topology explicitly on every node, you simply start a single node in bootstrap mode (using `--cluster_bootstrap`) and then when subsequent nodes are started, they will automatically join the cluster. Bootstrapping only takes effect if there is no previously stored cluster state.

@kozlovic


tylertreat:
One thing I want to change after this is merged is remove the need to configure a cluster node ID. Since the cluster automatically forms after starting a seed node, there shouldn't be a need to set this. We could just use the server ID instead (assuming that is stored/recovered from the store, which I need to check).

tylertreat:
It looks like server ID is generated on each restart. @kozlovic What are your thoughts on durably storing it for the purposes of clustering described above?

server/clustering.go
```diff
@@ -47,6 +51,9 @@ func (r *raftNode) shutdown() error {
 	if err := r.transport.Close(); err != nil {
 		return err
 	}
+	if err := r.joinSub.Unsubscribe(); err != nil {
```
kozlovic: I see that we create the `transport` based on `s.ncr`. I assume `r.transport.Close()` is not closing the underlying NATS Connection, right? If so, the unsubscribe here would fail.

kozlovic:
@tylertreat wrt server ID: I would rather that you add a new field, say NodeID used by clustering. The reason I would not want serverID to be persisted is that for instance in some situations, users would use different serverID in monitoring endpoints as a detection that server was restarted. Making it persistent would break such assumption.

server/clustering.go
tylertreat: Your assumption is correct. Transport close here just shuts down the Raft transport but not the underlying NATS conn.

PR #423, opened 2017-11-06
Make cluster node id flag optional and add clustered flag

tylertreat:
Generate a unique cluster node id if not set and store/recover this on
startup. Add a new `--clustered` flag users must set to indicate running
in clustered configuration.

@kozlovic


server/server.go
```diff
@@ -1558,6 +1555,9 @@ func (s *StanServer) start(runningState State) error {
 			callStoreInit = true
 		}
 
+		// Use recovered clustering node ID.
+		s.opts.Clustering.NodeID = s.info.NodeID
```
kozlovic: You may want to either bail out or support the case where user starts in clustering mode without NodeID in options and the store does not contain a NodeID (either from an older release or from runs with new version but without clustering mode and then running in clustering mode).

tylertreat: It's okay to start server in clustered mode without providing node id since a random one is assigned otherwise.

kozlovic: Don't you set it to a nuid in the `else` of this if?

kozlovic: Do we plan to support migrating from existing store to cluster mode? If we do, then you should do as for `SubClose` case just above: that is, the proto field was not set because it did not exist. So you detect that value is not set and was not provided from command line, so assign a new one and set `callStoreInit` to true (so that it gets updated in storage).
Thinking about it, it is not just a matter of support from older to newer release: situation would be the same if you started new server (cluster capable) in standalone mode and then restarted with clustering enabled: the nodeID would not be present.

kozlovic: so the question is more: do we want to support starting a server that was previously started as standalone to cluster mode.

kozlovic: it may actually make no sense because the corresponding raft logs would be missing: there would be things in streaming stores, but nothing in raft logs. So it may make the system inconsistent. So maybe we need to have a way that says: you can't start with existing store files if  you start in cluster mode and this field is not present?

tylertreat: Yeah, I think it would be problematic to restart a previously running server in clustered mode for the reasons you outline. Would it be sufficient to simply panic with a detailed error message when there is previously stored state and no cluster node id? Then only assign a node id if clustering is enabled.

kozlovic: You can actually return an error. The `RunServerWithOpts` returns an instance and error, so users running NATS Streaming programmatically would be able to check, and main.go will also report the error. It is a bit nicer than the panic.

PR #427, opened 2017-11-10
Implement channel gossiping

tylertreat:
Have servers periodically gossip their known channels to the cluster and
allow servers to reconcile their channels.

@kozlovic


server/server.go
```diff
@@ -1656,6 +1665,106 @@ func (s *StanServer) start(runningState State) error {
...
+		}
+	})
+	if err != nil {
+		return err
+	}
+
+	// Setup sub for receiving channel requests.
+	reqInbox := fmt.Sprintf("%s.%s.channels.%s", defaultGossipPrefix, s.opts.ID, s.opts.Clustering.NodeID)
+	_, err = s.ncr.Subscribe(reqInbox, func(m *nats.Msg) {
+		s.channels.RLock()
```
kozlovic: You may want to create a `channelStore.getAllNames()` or something to do that. We have a `channelStore.getAll()` that returns a map of all channels, but I understand that you want an array, and only of the names.

server/server.go
```diff
@@ -1656,6 +1665,106 @@ func (s *StanServer) start(runningState State) error {
...
+		channels := make([]string, len(s.channels.channels))
+		i := 0
+		for channel, _ := range s.channels.channels {
+			channels[i] = channel
+			i++
+		}
+		s.channels.RUnlock()
+		resp, err := (&spb.ChannelResponse{Channels: channels}).Marshal()
```
kozlovic: You can't do that since the list may be too big to fit in a NATS message. I have done splitting of channels lists for partitioning (check `sendChannelsList()` in `partitions.go`), but the code would need to be made more generic (like accept a connection, subject, list of channels, etc...).

server/server.go
```diff
@@ -1656,6 +1665,106 @@ func (s *StanServer) start(runningState State) error {
...
+			panic(err)
+		}
+		s.ncr.Publish(m.Reply, resp)
+	})
+	if err != nil {
+		return err
+	}
+
+	// Start publishing channel info.
+	go func() {
```
kozlovic: A bit late saying that, but with clustering we started to use quite a bit of go routines, and I would have like to make sure that we wait on them when server shutdown (more for tests than anything). In this case, you should call s.wg.Add(1) before the `go func()` and start the func with a `defer s.wg.Done()`.
When you have situations where you want to start a server go routine, and you are not currently holding the `s.mu.Lock()`, then you can use `StanServer.startGoRoutine()` which ensures that the server is not being shutdown, and if so, bumps the `s.wg` count and starts the given go routine.

tylertreat: This is what the `shutdownCh` is for though? Or are you saying it's not sufficient to have the goroutine eventually exit but instead have shutdown only return once we know the goroutine has exited?

kozlovic: Yes, the go routine will eventually stop, but sometimes, and again essentially for tests, it is better to ensure that things are done before moving to the next test, so yes, the idea is to have Shutdown() not return until go routines are done. It may not be an issue with this go routine though.

kozlovic: .. but good practice as we start go routines.

kozlovic: This is not a blocker, though. The splitting of the list to fit into NATS messages is.

tylertreat:
@kozlovic I've addressed the channel chunking by reusing some of the logic from partitioning. I will address goroutine shutdown in a future PR.

server/server.go
```diff
@@ -1656,6 +1665,145 @@ func (s *StanServer) start(runningState State) error {
...
+	if len(m.Data) == 0 {
+		return
+	}
+	req := spb.CtrlMsg{}
+	if err := req.Unmarshal(m.Data); err != nil {
+		s.log.Errorf("Error processing channel reconcile request: %v", err)
+		return
+	}
+	// Ignore messages from ourselves.
+	if req.ServerID == s.serverID {
```
kozlovic: Looks like this could not happen, right? A server is receiving the list on an private subject and it would not have send the request to itself.

kozlovic: In partitioning code, the subject was public, so a server would have to exclude control messages that it received from itself.

server/server.go
```diff
@@ -1656,6 +1665,145 @@ func (s *StanServer) start(runningState State) error {
...
+		return
+	}
+	channels, err := util.DecodeChannels(req.Data)
```
kozlovic: It is understood that this is not the complete list. SendChannelLists() sends possible multiple NATS messages, so this current callback may be invoked multiple times for the same "list".

server/server.go
tylertreat: Yes, I will remove the check as it's unnecessary.

server/server.go
tylertreat: Right, the server will send the complete list in chunks so this will get invoked for each chunks. It's possible this could get called redundantly if multiple servers are streaming their channel list, but this is OK since channel reconciliation is idempotent (and for now, we don't need to worry about channel deletion).

PR #428, opened 2017-11-14
Support manual bootstrap configuration of cluster

tylertreat:
Now clusters can be bootstrapped either by starting a node in bootstrap
mode and allowing other nodes to automatically join or by starting
servers with a cluster configuration. Once the cluster is formed, new
nodes can join the cluster automatically.

@kozlovic

### DIFF
--- a/server/clustering.go
+++ b/server/clustering.go
@@ -257,32 +257,25 @@ func (s *StanServer) createRaftNode(name string, fsm raft.FSM) (*raftNode, error
 // and with the latter taking precedence.
 func (s *StanServer) bootstrapCluster(name string, node *raft.Raft) error {
 	var (
-		servers []raft.Server
-		addr    = s.getClusteringAddr(name)
+		addr = s.getClusteringAddr(name)
+		// Include ourself in the cluster.
+		servers = []raft.Server{raft.Server{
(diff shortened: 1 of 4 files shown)


server/clustering.go
```diff
@@ -256,6 +252,46 @@ func (s *StanServer) createRaftNode(name string, fsm raft.FSM) (*raftNode, error
 	}, nil
 }
 
+// bootstrapCluster bootstraps the node for the provided Raft group either as a
+// seed node or with the given peer configuration, depending on configuration
+// and with the latter taking precedence.
+func (s *StanServer) bootstrapCluster(name string, node *raft.Raft) error {
```
kozlovic: Something look strange here. Since we get in this function with the guarantee that both Bootstrap can't be false and Peers be empty, In both cases you add the current node as the first element. There is no indication when calling `node.BootstrapCluster()` that we are in a bootstrap (seed) mode or not. So if the current code is correct, it looks like you could first do an append of the current node, then if `len(s.opts.Clustering.Peers) > 0` append the peers.
Am I missing something here?

tylertreat: Yes, you're right. This was mostly an artifact of some refactoring I did. I will simplify.

PR #438, opened 2017-11-29
[ADDED] Support for FT with SQL Store

kozlovic:
Simulate locking of database with use of table with single record
that is updated with owner (nuid) and tick. The table is updated
at regular interval. If owner is empty, lock can be acquired right
away. If not, instance checks if tick is updated after waiting for
1.5*interval. If tick has changed, lock is assumed healthy.
If after a configurable amount of non updates, an instance can
steal the lock.
If the owner of the lock fails to update the lock table, the
process will panic.


stores/sqlstore.go
```diff
@@ -359,14 +391,156 @@ func NewSQLStore(log logger.Logger, driver, source string, limits *StoreLimits,
...
+		return false, "", 0, sqlStmtError(sqlDBLockSelect, err)
+	}
+	if err == sql.ErrNoRows || steal || lockID == "" || lockID == s.dbLock.id {
+		// If we are stealing, reset tick to 0 (so it will become 1 in update statement)
+		if steal {
+			tick = 0
+		}
+		stmt := sqlStmts[sqlDBLockUpdate]
+		if err == sql.ErrNoRows {
+			stmt = sqlStmts[sqlDBLockInsert]
```
petemiron: I really want to believe this doesn't work this way, but tested it directly in MySQL and it does! Thought `FOR UPDATE` only locked selected rows.

kozlovic: Yes, I first I required to have at least a row in that table (even when creating the table, I had an insert with blank values in there). But then I realized that it would work in that if there is 2 selects for update when the table is empty, one of them will fail at the time of commit, so that solves the issue.

PR #445, opened 2017-12-08
[FIXED] Possible race between connect and pub/sub with partitioning

kozlovic:
In partitioning mode, when a client connects, the connect request
may reach several servers, but the first response the client gets
allows it to proceed with either publish or subscribe.
So it is possible for a server running in partitioning mode to
receives a connection request followed by a message or subscription.
Although the conn request would be first in the tcp connection, it
is then possible that the PubMsg or SubscriptionRequest be processed
first due to the use of different nats subscriptions.
To prevent that, when checking if a client exists, in this particular
mode we will possibly wait to be notified when the client has been
registered.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3,16 +3,18 @@
 package server
 
 import (
-	"github.com/nats-io/nats-streaming-server/stores"
 	"sync"
 	"time"
+
+	"github.com/nats-io/nats-streaming-server/stores"
 )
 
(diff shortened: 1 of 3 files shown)


kozlovic:
@tylertreat That may also solve the race that we have in clustering.. would need to check that.

tylertreat:
I think this would fix the race in clustering as well.

PR #458, opened 2018-01-22
[FIXED] FT Standby may fail to start in Channels Partitioning mode

### LINKED ISSUE #453: Partitioning ft_group on 3 kubernetes nodes
This works on a stateful 3 node kubernetes set where each node gets a stream pod, but it seems when enabling partitions you can only have 1 active and 1 standby instead of 1 active and 2 standby (one on each node) The 3 one will crash with rpc limit reached?
```
https: 0.0.0.0:8222
listen: 0.0.0.0:4222
tls {
  cert_file: "/etc/nats/tls/nats.pem"
  key_file:  "/etc/nats/tls/nats-key.pem"
  ca_file:   "/etc/nats/tls/ca.pem"
  verify:   true
}
authorization {
  user:     nats_client_user
  password: nats_client_pwd
}
debug:   true
trace:   true
logtime: true
max_connections: 100
max_control_line: 512
max_payload: 65536
write_deadline: "2s"

cluster {
  listen: 0.0.0.0:6222
  tls {
    cert_file:  "/etc/nats/tls/nats.pem"
    key_file:   "/etc/nats/tls/nats-key.pem"
    ca_file:    "/etc/nats/tls/ca.pem"
  }
  authorization {
    user: nats_route_user
    password: nats_route_pwd
  }
  routes = [
    nats://nats_route_user:nats_route_pwd@nats:6222
  ]
}

streaming {
  cluster_id: nats-streaming
  nats_server_url: nats://nats_client_user:nats_client_pwd@nats:4222
  store: FILE # MEMORY
  dir: "/store"
  ft_group: "ftg"
  secure: true
  tls {
    client_cert: "/etc/nats/tls/nats.pem"
    client_key:  "/etc/nats/tls/nats-key.pem"
    client_ca:   "/etc/nats/tls/ca.pem"
  }
  partitioning: false
  store_limits {
    channels {
      bench-subject {}
    }
  }
}
```
Ps feel free to comment on my config if you see mistakes, thx
Also I think that having a `client {...}` is going to be less confusing so you can have 3 config blocks client cluster and streaming

kozlovic:
In Partitioning mode, when a server starts, it tries to see if
another server in the NATS cluster has incorrectly configured the
same channel(s). However, an FT server should do this test only
when it activates.

Resolves #453

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -33,7 +33,7 @@ import (
 // Server defaults.
 const (
 	// VERSION is the current version for the NATS Streaming server.
-	VERSION = "0.7.0"
+	VERSION = "0.7.1"
 
 	DefaultClusterID      = "test-cluster"
 	DefaultDiscoverPrefix = "_STAN.discover"
@@ -325,6 +325,7 @@ type StanServer struct {
 	info       spb.ServerInfo // Contains cluster ID and subjects
 	natsServer *server.Server
 	opts       *Options
+	natsOpts   *server.Options
 	startTime  time.Time
 
 	// For scalability, a dedicated connection is used to publish
@@ -907,9 +908,10 @@ func (s *StanServer) stanErrorHandler(nc *nats.Conn, sub *nats.Subscription, err
 		nc.Opts.Name, sub.Subject, err)
 }
 
-func (s *StanServer) buildServerURLs(sOpts *Options, opts *server.Options) ([]string, error) {
+func (s *StanServer) buildServerURLs() ([]string, error) {
 	var hostport string
-	natsURL := sOpts.NATSServerURL
+	natsURL := s.opts.NATSServerURL
+	opts := s.natsOpts
 	// If the URL to an external NATS is provided...
 	if natsURL != "" {
 		// If it has user/pwd info or is a list of urls...
@@ -964,15 +966,15 @@ func (s *StanServer) buildServerURLs(sOpts *Options, opts *server.Options) ([]st
 // createNatsClientConn creates a connection to the NATS server, using
 // TLS if configured.  Pass in the NATS server options to derive a
 // connection url, and for other future items (e.g. auth)
-func (s *StanServer) createNatsClientConn(name string, sOpts *Options, nOpts *server.Options) (*nats.Conn, error) {
+func (s *StanServer) createNatsClientConn(name string) (*nats.Conn, error) {
 	var err error
 	ncOpts := nats.DefaultOptions
 
-	ncOpts.Servers, err = s.buildServerURLs(sOpts, nOpts)
+	ncOpts.Servers, err = s.buildServerURLs()
 	if err != nil {
 		return nil, err
 	}
-	ncOpts.Name = fmt.Sprintf("_NSS-%s-%s", sOpts.ID, name)
+	ncOpts.Name = fmt.Sprintf("_NSS-%s-%s", s.opts.ID, name)
 
 	if err = nats.ErrorHandler(s.stanErrorHandler)(&ncOpts); err != nil {
 		return nil, err
@@ -986,18 +988,18 @@ func (s *StanServer) createNatsClientConn(name string, sOpts *Options, nOpts *se
 	if err = nats.DisconnectHandler(s.stanDisconnectedHandler)(&ncOpts); err != nil {
 		return nil, err
 	}
-	if sOpts.Secure {
+	if s.opts.Secure {
 		if err = nats.Secure()(&ncOpts); err != nil {
 			return nil, err
 		}
 	}
-	if sOpts.ClientCA != "" {
-		if err = nats.RootCAs(sOpts.ClientCA)(&ncOpts); err != nil {
+	if s.opts.ClientCA != "" {
+		if err = nats.RootCAs(s.opts.ClientCA)(&ncOpts); err != nil {
 			return nil, err
 		}
 	}
-	if sOpts.ClientCert != "" {
-		if err = nats.ClientCert(sOpts.ClientCert, sOpts.ClientKey)(&ncOpts); err != nil {
+	if s.opts.ClientCert != "" {
+		if err = nats.ClientCert(s.opts.ClientCert, s.opts.ClientKey)(&ncOpts); err != nil {
 			return nil, err
 		}
 	}
@@ -1022,14 +1024,14 @@ func (s *StanServer) createNatsClientConn(name string, sOpts *Options, nOpts *se
 	return nc, err
 }
 
-func (s *StanServer) createNatsConnections(sOpts *Options, nOpts *server.Options) error {
+func (s *StanServer) createNatsConnections() error {
 	var err error
-	s.ncs, err = s.createNatsClientConn("send", sOpts, nOpts)
+	s.ncs, err = s.createNatsClientConn("send")
 	if err == nil {
-		s.nc, err = s.createNatsClientConn("general", sOpts, nOpts)
+		s.nc, err = s.createNatsClientConn("general")
 	}
-	if err == nil && sOpts.FTGroupName != "" {
-		s.ftnc, err = s.createNatsClientConn("ft", sOpts, nOpts)
+	if err == nil && s.opts.FTGroupName != "" {
+		s.ftnc, err = s.createNatsClientConn("ft")
 	}
 	return err
 }
@@ -1062,6 +1064,7 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 	s := StanServer{
 		serverID:          nuid.Next(),
 		opts:              sOpts,
+		natsOpts:          nOpts,
 		dupCIDMap:         make(map[string]struct{}),
 		dupMaxCIDRoutines: defaultMaxDupCIDRoutines,
 		dupCIDTimeout:     defaultCheckDupCIDTimeout,
@@ -1081,7 +1084,7 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 	if sOpts.CustomLogger != nil {
 		s.log.SetLogger(sOpts.CustomLogger, sOpts.Debug, sOpts.Trace)
 	} else if sOpts.EnableLogging {
-		s.configureLogger(nOpts)
+		s.configureLogger()
 	}
 
 	s.log.Noticef("Starting nats-streaming-server[%s] version %s", sOpts.ID, VERSION)
@@ -1162,7 +1165,7 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 
 	// If no NATS server url is provided, it means that we embed the NATS Server
 	if sOpts.NATSServerURL == "" {
-		if err := s.startNATSServer(nOpts); err != nil {
+		if err := s.startNATSServer(); err != nil {
 			return nil, err
 		}
 	}
@@ -1173,16 +1176,8 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 		}
 	}
 	// Create our connections
-	if err := s.createNatsConnections(sOpts, nOpts); err != nil {
+	if err := s.createNatsConnections(); err != nil {
 		return nil, err
-	}
-
-	// If using partitioning, try our best to find out that on startup that
-	// no other server with same cluster ID has any channel that we own.
-	if sOpts.Partitioning {
-		if err := s.initPartitions(sOpts, nOpts, storeLimits.PerChannel); err != nil {
-			return nil, err
-		}
 	}
 
 	// In FT mode, server cannot recover the store until it is elected leader.
@@ -1218,10 +1213,11 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 // trace/debug if the local STAN debug/trace flags are set.  NATS will do
 // the same with it's logger flags.  This enables us to use the same logger,
 // but differentiate between STAN and NATS debug/trace.
-func (s *StanServer) configureLogger(nOpts *server.Options) {
+func (s *StanServer) configureLogger() {
 	var newLogger logger.Logger
 
 	sOpts := s.opts
+	nOpts := s.natsOpts
 
 	enableDebug := nOpts.Debug || sOpts.Debug
 	enableTrace := nOpts.Trace || sOpts.Trace
@@ -1257,6 +1253,14 @@ func (s *StanServer) start(runningState State) error {
 	defer s.mu.Unlock()
 	if s.shutdown {
 		return nil
+	}
+
+	// If using partitioning, send our list and start go routines handling
+	// channels list requests.
+	if s.opts.Partitioning {
+		if err := s.initPartitions(); err != nil {
+			return err
+		}
 	}
 
 	var (
@@ -1378,7 +1382,8 @@ func (s *StanServer) start(runningState State) error {
 }
 
 // TODO:  Explore parameter passing in gnatsd.  Keep separate for now.
-func (s *StanServer) configureClusterOpts(opts *server.Options) error {
+func (s *StanServer) configureClusterOpts() error {
+	opts := s.natsOpts
 	// If we don't have cluster defined in the configuration
 	// file and no cluster listen string override, but we do
 	// have a routes override, we need to report misconfiguration.
@@ -1438,7 +1443,8 @@ func (s *StanServer) configureClusterOpts(opts *server.Options) error {
 // configureNATSServerTLS sets up TLS for the NATS Server.
 // Additional TLS parameters (e.g. cipher suites) will need to be placed
 // in a configuration file specified through the -config parameter.
-func (s *StanServer) configureNATSServerTLS(opts *server.Options) error {
+func (s *StanServer) configureNATSServerTLS() error {
+	opts := s.natsOpts
 	tlsSet := false
 	tc := server.TLSConfigOpts{}
 	if opts.TLSCert != "" {
@@ -1469,15 +1475,16 @@ func (s *StanServer) configureNATSServerTLS(opts *server.Options) error {
 	return nil
 }
 
-// startNATSServer massages options as necessary, and starts the embedded
-// NATS server.
-func (s *StanServer) startNATSServer(opts *server.Options) error {
-	if err := s.configureClusterOpts(opts); err != nil {
+// startNATSServer starts the embedded NATS server, possibly updating
+// the NATS Server's clustering and/or TLS options.
+func (s *StanServer) startNATSServer() error {
+	if err := s.configureClusterOpts(); err != nil {
 		return err
 	}
-	if err := s.configureNATSServerTLS(opts); err != nil {
+	if err := s.configureNATSServerTLS(); err != nil {
 		return err
 	}
+	opts := s.natsOpts
 	s.natsServer = server.New(opts)
 	if s.natsServer == nil {
 		return fmt.Errorf("no NATS Server object returned")
--- a/server/partitions.go
+++ b/server/partitions.go
@@ -7,7 +7,6 @@ import (
 	"sync"
 	"time"
 
-	natsd "github.com/nats-io/gnatsd/server"
 	"github.com/nats-io/go-nats"
 	"github.com/nats-io/nats-streaming-server/spb"
 	"github.com/nats-io/nats-streaming-server/stores"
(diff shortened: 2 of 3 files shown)


server/server.go
```diff
@@ -1471,13 +1477,14 @@ func (s *StanServer) configureNATSServerTLS(opts *server.Options) error {
 
 // startNATSServer massages options as necessary, and starts the embedded
```
gertcuykens: typo messages i think

kozlovic: It may be bad wording, but I think it meant that the nats options are changed if needed in order to start the embedded NATS Server.

gertcuykens: Ok thx suggest `startNATSServer change options if necessary,`

kozlovic: What about:
```
// startNATSServer starts the embedded NATS server, possibly updating
// the NATS Server's clustering and/or TLS options.
```
or remove altogether ;-)

kozlovic:
@gertcuykens Thanks for the review. Do you think you will be able to test this in your setup? Thanks!

gertcuykens:
I think so tommorow

gertcuykens:
Hmm dont think this is the docker file i need to build
https://github.com/nats-io/nats-streaming-server/blob/master/Dockerfile

Can you give me a docker image or Dockerfile i can use please to replace my current yaml
```
      containers:
      - name: stan
        image: nats-streaming:0.7.0
        args: [ "--config", "/etc/nats/nats.conf" ]
```

kozlovic:
No, you would need to clone `https://github.com/nats-io/nats-streaming-docker` then from there, do a `./update.sh fix_ft_and_channel_partitioning` to update the executables in that repo to the code from the branch. Then, from the `amd64` directory, do a `docker build -t test .` which would build the image `test`.

kozlovic:
@gertcuykens By the way, if you can't test it, that's fine. Just let me know.

gertcuykens:
hmm i now get 
```
[1] 2018/01/23 03:20:42.509304 [INF] STREAM: Starting nats-streaming-server[nats-streaming] version 0.7.0
[1] 2018/01/23 03:20:42.509710 [INF] STREAM: ServerID: Uo1ZYEfIgwcDbpgeOFvCPF
[1] 2018/01/23 03:20:42.509734 [INF] STREAM: Go version: go1.9.2
[1] 2018/01/23 03:20:42.773283 [INF] STREAM: Starting in standby mode
[1] 2018/01/23 03:20:44.031839 [INF] STREAM: Server is active
[1] 2018/01/23 03:20:44.032101 [FTL] STREAM: nats: error loading or parsing rootCA file: open /etc/nats/tls/ca.pem: no such file or directory
```
But only changed the image and the certificates are in the same place, and i can still see them in my kubernetes secrets

gertcuykens:
when i switch back to version 0.7.0 it works again and finds my certificates
```
---
apiVersion: v1
kind: Service
metadata:
  name: stan
  labels:
    component: stan
spec:
  selector:
    component: stan
  clusterIP: None
  ports:
  - name: client
    port: 4222
  - name: cluster
    port: 6222
  - name: monitor
    port: 8222

---
apiVersion: apps/v1beta1
kind: StatefulSet
metadata:
  name: stan
  labels:
    component: stan
spec:
  serviceName: stan
  replicas: 3
  template:
    metadata:
      labels:
        component: stan
    spec:
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: component
                operator: In
                values:
                - stan
            topologyKey: kubernetes.io/hostname
      containers:
      - name: stan
        image: nats-streaming:0.7.0
        # image: xxxxxx:5000/nats-streaming-server
        args: [ "--config", "/etc/nats/nats.conf" ]
        volumeMounts:
        - name: tls-volume
          mountPath: /etc/nats/tls
        - name: config-volume
          mountPath: /etc/nats
        - name: store-volume
          mountPath: /store
        ports:
        - containerPort: 4222
          name: client
        - containerPort: 6222
          name: cluster
        - containerPort: 8222
          name: monitor
        livenessProbe:
          httpGet:
            path: /
            port: 8222
            scheme: HTTPS
          initialDelaySeconds: 10
          timeoutSeconds: 5
      volumes:
      - name: tls-volume
        secret:
          secretName: tls-nats-server
      - name: config-volume
        configMap:
          name: nats-config
      - name: store-volume
        emptyDir:
          medium: Memory
```

gertcuykens:
Could it be it tries to start embedded mode instead of using the nats cluster or try to use one of the other certificates now?

kozlovic:
@gertcuykens I wonder if this is instead due to the way you reference the custom image.

If you want, you could build another image the way you did it before, but from the `nats-streaming-docker` root repo, do `./update master` instead, then from `amd64` build a new image. Try with that one. This would have the bug about the channel detection, but it would allow us to see if you have the same issue with certificate.

There are only 2 commits that separate master from the official docker image. If you still have the cert issue, then from the `amd64` directory, do a `git checkout .` to revert to original executable and build again your test image (which should be equivalent to the official docker image).

If there is still the certificate issue, it means that this is a problem the way the custom image is executed.

gertcuykens:
master is working, fix_ft_and_channel_partitioning result in can't find certificate error if partitioning: true, when false no certificate error and working ?

gertcuykens:
In addition when partitioning is false i am able to start 2 at the same time now using  fix_ft_and_channel_partitioning
image
image
image

kozlovic:
I updated the PR but it should not make a difference (removed unused parameters from a function). I am a bit at loss. I spent most of today trying to reproduce or understand what is going on, and I don't see how the changes would explain what you are seeing.
I believe that this is the proper fix: I have tested locally and works well. I even deployed in a k8s cluster (but without TLS to be honest).

I suspect that there is something wrong with your deployment. For instance, with 0.7.0, it should NOT work at all, that is, you should not be able to run more than 1 instance. Yet, in some of the outputs, we could see that 1 of the 3 instances was reporting the failure, but one was saying that it was in standby mode. It should not be if it had connectivity with the cluster and the other STAN instance running. That is, it should have detected another instance with the same channel (due to the bug) and fail to start.

This may also explain why you said that with the image from the branch you can get 2 active instances. It seem to indicate that there is no connectivity and in this case even not sharing of the same datastore.

As for the error regarding the loading of the certificate, it shows that the options are correctly set (we see that it tried to load `/etc/nats/tls/ca.pem` but this file is not accessible from the test image. The risk with the changes in this PR would have been that the options that are now accessed through the `StanServer` struct would somehow be different than when they were accessed directly in `RunServerWithOpts()`, but this is not the case.

If you could provide charts and or config files allowing us to set our env here and try to reproduce that would be great. Otherwise, I may produce a 0.7.2 release later this week or beginning of next.

gertcuykens:
stan.yaml
```
---
apiVersion: v1
kind: Service
metadata:
  name: stan
  labels:
    component: stan
spec:
  selector:
    component: stan
  clusterIP: None
  ports:
  - name: client
    port: 4222
  - name: cluster
    port: 6222
  - name: monitor
    port: 8222

---
apiVersion: apps/v1beta1
kind: StatefulSet
metadata:
  name: stan
  labels:
    component: stan
spec:
  serviceName: stan
  replicas: 3
  template:
    metadata:
      labels:
        component: stan
    spec:
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: component
                operator: In
                values:
                - stan
            topologyKey: kubernetes.io/hostname
      containers:
      - name: stan
        # image: nats-streaming:0.7.0
        image: repository.be-mobile.biz:5000/nats-streaming-server:latest
        args: [ "--config", "/etc/nats/nats.conf" ]
        volumeMounts:
        - name: tls-volume
          mountPath: /etc/nats/tls
        - name: config-volume
          mountPath: /etc/nats
        - name: store-volume
          mountPath: /store
        ports:
        - containerPort: 4222
          name: client
        - containerPort: 6222
          name: cluster
        - containerPort: 8222
          name: monitor
        livenessProbe:
          httpGet:
            path: /
            port: 8222
            scheme: HTTPS
          initialDelaySeconds: 10
          timeoutSeconds: 5
      volumes:
      - name: tls-volume
        secret:
          secretName: tls-nats-server
      - name: config-volume
        configMap:
          name: nats-config
      - name: store-volume
        emptyDir:
          medium: Memory
```
nats.yaml
```
---
apiVersion: v1
kind: Service
metadata:
  name: nats
  labels:
    component: nats
spec:
  selector:
    component: nats
  clusterIP: None
  ports:
  - name: client
    port: 4222
  - name: cluster
    port: 6222
  - name: monitor
    port: 8222

---
apiVersion: apps/v1beta1
kind: StatefulSet
metadata:
  name: nats
  labels:
    component: nats
spec:
  serviceName: nats
  replicas: 3
  template:
    metadata:
      labels:
        component: nats
    spec:
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: component
                operator: In
                values:
                - nats
            topologyKey: kubernetes.io/hostname
      containers:
      - name: nats
        image: nats:1.0.4
        args: [ "--config", "/etc/nats/nats.conf"]
        volumeMounts:
        - name: tls-volume
          mountPath: /etc/nats/tls
        - name: config-volume
          mountPath: /etc/nats
        ports:
        - containerPort: 4222
          name: client
        - containerPort: 6222
          name: cluster
        - containerPort: 8222
          name: monitor
        livenessProbe:
          httpGet:
            path: /
            port: 8222
            scheme: HTTPS
          initialDelaySeconds: 10
          timeoutSeconds: 5
      volumes:
      - name: tls-volume
        secret:
          secretName: tls-nats-server
      - name: config-volume
        configMap:
          name: nats-config
```
nats.conf
```
https: 0.0.0.0:8222
listen: 0.0.0.0:4222
tls {
  cert_file: "/etc/nats/tls/nats.pem"
  key_file:  "/etc/nats/tls/nats-key.pem"
  ca_file:   "/etc/nats/tls/ca.pem"
  verify:   false
}
authorization {
  user:     nats_client_user
  password: nats_client_pwd
}
debug:   true
trace:   true
logtime: true
max_connections: 100
max_control_line: 512
max_payload: 65536
write_deadline: "2s"

cluster {
  listen: 0.0.0.0:6222
  tls {
    cert_file:  "/etc/nats/tls/nats.pem"
    key_file:   "/etc/nats/tls/nats-key.pem"
    ca_file:    "/etc/nats/tls/ca.pem"
  }
  authorization {
    user: nats_route_user
    password: nats_route_pwd
  }
  routes = [
    nats://nats_route_user:nats_route_pwd@nats:6222
  ]
}

streaming {
  cluster_id: nats-streaming
  nats_server_url: nats://nats_client_user:nats_client_pwd@nats:4222
  store: FILE # MEMORY
  dir: "/store"
  ft_group: "ft"
  secure: true
  tls {
    client_cert: "/etc/nats/tls/nats.pem"
    client_key:  "/etc/nats/tls/nats-key.pem"
    client_ca:   "/etc/nats/tls/ca.pem"
  }
  partitioning: false
  store_limits {
    channels {
      bench-subject {}
    }
  }
}
```
Makefile
```
tls:
	openssl genrsa -out tls/ca-key.pem 2048
	openssl req -x509 -new -nodes -key tls/ca-key.pem -days 10000 -out tls/ca.pem -subj "/CN=kube-ca"
	openssl genrsa -out tls/nats-key.pem 2048
	openssl req -new -key tls/nats-key.pem -out tls/nats.csr -subj "/CN=kube-nats" -config tls/tls.cnf
	openssl x509 -req -in tls/nats.csr -CA tls/ca.pem -CAkey tls/ca-key.pem -CAcreateserial -out tls/nats.pem -days 3650 -extensions v3_req -extfile tls/tls.cnf

kubeTLS:
	-kubectl delete secret tls-nats-client
	kubectl create secret generic tls-nats-client --from-file tls/ca.pem
	-kubectl delete secret tls-nats-server
	kubectl create secret generic tls-nats-server --from-file tls/nats.pem --from-file tls/nats-key.pem --from-file tls/ca.pem

kubeCFG:
	-kubectl delete configmap nats-config
	kubectl create configmap nats-config --from-file nats.conf

kubeNTS: kubeCFG
	-kubectl delete -f nats.yaml
	kubectl create -f nats.yaml

kubeSTN: kubeCFG
	-kubectl delete -f stan.yaml
	kubectl create -f stan.yaml

# kubectl logs -f nats-0
# kubectl scale statefulsets nats --replicas 5
# kubectl port-forward nats-0 8222:8222
# kubectl exec -it nats-0 -- /gnatsd -sl quit=1

.PHONY: tls
```
tls/tls.cnf
```
[req]
req_extensions = v3_req
distinguished_name = req_distinguished_name

[req_distinguished_name]

[v3_req]
basicConstraints = CA:FALSE
keyUsage = nonRepudiation, digitalSignature, keyEncipherment
subjectAltName = @alt_names

[alt_names]
DNS.1 = nats
IP.1 = 127.0.0.1
```

kozlovic:
Thanks! From what I understand, you are using the same config file for gnatsd and nats-streaming-server, right? As I mentioned earlier, the face that you specify `nats_server_url` means that the streaming server will not attempt to start an embedded instance but connect to the given url.

gertcuykens:
Correct I have a 3 node cluster running nats so i don't need the embedded server, also use a custom test pod to make sure nats_server_url: nats://nats_client_user:nats_client_pwd@nats:4222 is working

image

gertcuykens:
De above yamls should reproduce nats-0 nats-1 nats-2 (nats://nats_client_user:nats_client_pwd@nats:4222) and a stan-1 stan-2 stan-3

kozlovic:
Thanks! Will try to reproduce this likely tomorrow. I see that there are 3 stan running (with partitioning disabled) and the image built from the branch. Is only one of the pod active and 2 standbys or do you still have a case where more than 1 pod is active?

gertcuykens:
As far as I can recall `fix_ft_and_channel_partitioning` is the only one i was able to run more then one active stan server but that could be coincidence as in a race condition between terminating and starting pods. I think their is a small time window where it's possible to fool a stan client server. I understand this is very hard to debug but kubernetes is worth the effort to unlock massive auto horizontal scaling, Thanks :)

kozlovic:
@gertcuykens Thank you so much for sending the config files, it has been very helpful.

I have built the branch from the minikube docker environment and had to tweak the 2 yaml files, commenting out the affinity spec - otherwise it would not start, and replace the stan image with this:
```
        image: nss-test:latest
        imagePullPolicy: IfNotPresent
```

It worked, even when partitioning is enabled, I do see that some of the pods required a restart, indeed due to issue with the certificates. 

I added debug code to my local image and noticed that `config` and `store` volumes were always ok, however, the `tls` volume was ok only at the very beginning and then start to always return "not found" after a second or two (I had a go-routine started from the main checking the volumes).

This would explain why it "works" with master/0.7.0. This is because in partition, I create an additional connection that used to be created at the same time than the other, but in this branch I had moved that to when the server activates. I have updated the branch to create the connection early but start the check only on activation.

Still, I believe that this is an issue (or a feature) of the secret store that is not available after a short period of time. I think that this may be problematic and you may want to explore that.

The other issue I noticed are route errors in nats-1 and nats-2, like this:
```
[1] 2018/01/25 21:21:53.149462 [DBG] Trying to connect to route on 172.17.0.7:6222
[1] 2018/01/25 21:21:53.149860 [DBG] 172.17.0.7:6222 - rid:2 - Route connection created
[1] 2018/01/25 21:21:53.149886 [DBG] 172.17.0.7:6222 - rid:2 - Starting TLS route client handshake
[1] 2018/01/25 21:21:53.164020 [DBG] 172.17.0.7:6222 - rid:2 - TLS route handshake error: x509: certificate is valid for 127.0.0.1, not 172.17.0.7
[1] 2018/01/25 21:21:53.164168 [TRC] 172.17.0.7:6222 - rid:2 - &lt;&lt;- [-ERR Secure Connection - TLS Required]
[1] 2018/01/25 21:21:53.164255 [DBG] 172.17.0.7:6222 - rid:2 - Router connection closed
```
```
[1] 2018/01/25 21:21:53.148867 [DBG] 172.17.0.5:6222 - rid:1 - Registering remote route "tjOM67XddfIKDNhW7OgZHp"
[1] 2018/01/25 21:21:53.149218 [DBG] 172.17.0.5:6222 - rid:1 - Route sent local subscriptions
[1] 2018/01/25 21:21:53.150062 [DBG] 172.17.0.6:57800 - rid:2 - Route connection created
[1] 2018/01/25 21:21:53.150162 [DBG] 172.17.0.6:57800 - rid:2 - Starting TLS route server handshake
[1] 2018/01/25 21:21:53.164104 [DBG] 172.17.0.6:57800 - rid:2 - TLS route handshake error: remote error: tls: bad certificate
[1] 2018/01/25 21:21:53.164382 [TRC] 172.17.0.6:57800 - rid:2 - &lt;&lt;- [-ERR Secure Connection - TLS Required]
[1] 2018/01/25 21:21:53.164404 [DBG] 172.17.0.6:57800 - rid:2 - Router connection closed
```
With `kubectl port-forward nats-0 8222:8222` I could see that only `nats-0` had 2 routes, the other 2 instances had only 1 route. NATS needs full mesh. So if you have nats-1 -> nats-0 <- nats-2, this is not going to work. If an application (or streaming server) connects to nats-1, it won't be able to communicate with an application (or server) connected to nats-2, because nats-0 is not going to forward it. It is a 1-hop. That may explain dual active cases that you have seen.

Hoping to ear from you. I believe that if the NATS deployment was working fine (full mesh) we would not have the streaming issues with dual active. Also, you should pursue the issue with the tls volume not being available past a certain duration once the pod is started. The updated branch may reduce the risk of this issue for now.

gertcuykens:
1) The secret volume I am going to make a bug report on kubernetes repo because that sounds like a issue more people will encounter sooner or later. Thanks, never would have found that myself :)

2) Conclusion i think we can assume routes = [nats://nats_route_user:nats_route_pwd@nats:6222] will always result in half mesh because nats-0 (auto selected by kube load balancer but could be any) will not share / update / notify it friends list when another friend joins the party. Because of how kubernetes load balances and how headless service is designed, this handshake to nats-0 need to be expanded to something more fancier to unlock the full potential of kubernetes right? Can you convince the nats folks please to see if its possible to have a deeper cluster handshake implementation?

gertcuykens:
1) Can you try this to be sure stan is not touching the certificates resulting in a delete?
```
---
apiVersion: v1
kind: List
items:
- apiVersion: v1
  kind: Secret
  data:
    secret: bXkgc2VjcmV0Cg==
  metadata:
    name: secrets
  type: Opaque
- apiVersion: v1
  kind: Pod
  metadata:
    name: busybox-pod
  spec:
    containers:
    - image: gcr.io/google_containers/busybox:latest
      name: busybox
      command: ["/bin/sh"]
      args: ["-c", "sleep 10; echo before; ls -l /secrets; touch /secrets/myfile; echo checking; ls -l /secrets; sleep 120; echo checking again; ls -l /secrets; while sleep 3600; do true; done"]
      volumeMounts:
      - mountPath: /secrets
        name: tls-volume
    restartPolicy: OnFailure
    volumes:
    - name: tls-volume
      secret:
        secretName: tls-nats-server
```
Ref: https://github.com/kubernetes/kubernetes/issues/58719

kozlovic:
@gertcuykens For the full mesh issue, NATS does auto-discovery and as you can see in the excerpt of the logs, `nats-1` tried to connect to `nats-2` to form the full mesh, but the issue is that the name in certificate was not what `nats-1` is expecting based on the URL it is connecting to (the url was `172.17.0.7` but certificate has `127.0.0.1`).

We may add a feature to NATS Server to be able to provide a "TLS expected hostname" for such situation.

For now, you may want to simply disable TLS between the clustered nodes and just keep it for the clients.

I will merge the PR since I believe it addresses the problem. The issue with secret store is not related to NATS Streaming. As I was explaining, having a loop checking for the *directory* `/etc/nats/tls` was showing that it was ok the first time, but then started to report that the directory was not found. That was even without trying to load any certificate.

kozlovic:
I still dug a bit more and if you change your config mount in `stan.yaml` to be `/nats` (and changed `args` to `args: [ "--config", "/nats/nats.conf" ]` then the problem is solved!!!

It seems that there is an issue if the volume mount has a path in common between the secret and another volume (you currently have `/etc/nats/tls` for secret and `/etc/nats` for config).

Here is output before change:
image
And here after changing config volume to `/nats`:
image

kozlovic:
@gertcuykens Since some changes in the stan.yaml make it work, I reverted some of the changes and now delay the creation of the connection only on activation, as it was in the original commit of this PR. FYI I did a `git push --force`.

gertcuykens:
Tested master. Confirming everything works now, thx 👍

PR #468, opened 2018-02-14
[FIXED] Failed subscription appears in client's list of subscriptions

### LINKED ISSUE #467: Subscription not removed from client after subscription error
If a client subscribes to a channel and there is an error, the subscription fails, but the entry is not removed from the client. To reproduce this, allow 2 clients to subscribe to channel "foo.bar" using the following configuration file:

`store_limits {
	channels: {
		"foo.bar": {
			max_subs: 1
		}
	}
}`

The bug should occur when the second client attempts to subscribe to the channel. This can be checked at the **clientsz** endpoint (/streaming/clientsz?subs=1)

kozlovic:
When the MaxSubscriptions limit is set, an application would correctly
get an error when trying to create a subscription more than the limit.
However, this subscription would still appear in the list of
subscriptions for that client (as seen in the monitor endpoints).

Resolves #467

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -3728,7 +3728,11 @@ func (s *StanServer) addSubscription(ss *subStore, sub *subState) error {
 		return fmt.Errorf("can't find clientID: %v", sub.ClientID)
 	}
 	// Store this subscription in subStore
-	return ss.Store(sub)
+	if err := ss.Store(sub); err != nil {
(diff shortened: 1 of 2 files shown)


AshfordN:
I have tested this PR and it does in fact resolve bug #467

kozlovic:
@AshfordN Thanks for the report and confirmation that the fix works!

PR #469, opened 2018-02-14
[ADDED] Automatic deletion of channels after a certain inactivity

### LINKED ISSUE #358: Proposal: Support Deletion of a Channel
Right now, it is impossible to delete a channel.

For FileStore, one can stop the server and delete the sub directory
corresponding to the channel, then restart the server. But this 
causes downtime.

A proposal would be to allow deletion of a channel with following
restrictions:
- The deletion of the channel would be done through the monitoring
endpoint, such as `https://localhost:8222/channelsz?channel=foo&delete=1`
- It would be restricted to HTTPS port.
- Request would be rejected by the server if there are existing
subscriptions attached to the channel.

kozlovic:
Added MaxInactivity as a new channel limit. By default, this limit
is set to unlimited. If the user sets this limit, then when
a channel does not have any subscription and no new message
for at least MaxInactivity duration, the server will delete such
channel.

Resolves #358


laugimethods:
Since the main purpose of this feature is to reduce the number of open channels to limit the amount of required ressources, I suggest to add another (optional) threshold:  a number of open channels from which the `MaxInactivity` parameter will be effective.
=> when NATS is not overloaded, do nothing.  Otherwise, start to delete channels (based on  `MaxInactivity`).

kozlovic:
Note: I did update (and push force) the PR to exclude offline durable (queue) subscriptions from the restrictions related to the deletion of a channel.

abbi-gaurav:
@kozlovic, If I understand correctly, In case there are durable subscriptions, the channel will not be deleted even if it exceeds `MaxInactivity`?
Would it be possible to make this behavior configurable (to apply the restriction for durable subscriptions or not)?

kozlovic:
@abbi-gaurav It will. Look at the comment prior to yours. I did update the PR so that offline durables are not preventing deletion of a channel. Only active subscriptions (durables or not) would.

PR #522, opened 2018-03-22
fixed: support for declaring own node_id in peers

### LINKED ISSUE #521: clustering: support own node id in closer peers
Right now in order to start a 3 node nats-streaming-server cluster you must do:

```
node0: --cluster_id a --cluster_peers b,c
node1: --cluster_id b --cluster_peers a,c
node2: --cluster_id c --cluster_peers a,b
```

Doing this doesn't work:

```
node0: --cluster_id a --cluster_peers a,b,c
node1: --cluster_id b --cluster_peers a,b,c
node2: --cluster_id c --cluster_peers a,b,c
```

This doesn't make life easy on a  kubernetes statefulset.

If nats-streaming was simply pruning its own ID from the `cluster_peers` list, it would be very easy to write

```
node0: --cluster_id $(HOSTNAME) --cluster_peers ns-0,ns-1,ns-2
node1: --cluster_id $(HOSTNAME) --cluster_peers ns-0,ns-1,ns-2
node2: --cluster_id $(HOSTNAME) --cluster_peers ns-0,ns-1,ns-2
```

primalmotion:
fixes #521

### DIFF
--- a/server/conf.go
+++ b/server/conf.go
@@ -588,7 +588,9 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 	if clusterPeers != "" {
 		sopts.Clustering.Peers = []string{}
 		for _, p := range strings.Split(clusterPeers, ",") {
-			sopts.Clustering.Peers = append(sopts.Clustering.Peers, strings.TrimSpace(p))
+			if p = strings.TrimSpace(p); p != sopts.Clustering.NodeID {
+				sopts.Clustering.Peers = append(sopts.Clustering.Peers, p)
+			}
 		}
 	}
 
--- a/server/conf_test.go
+++ b/server/conf_test.go
@@ -200,15 +200,21 @@ func TestParseConfig(t *testing.T) {
 	if !opts.Clustering.Bootstrap {
 		t.Fatal("Expected Bootstrap to be true, got false")
 	}
-	peers := []string{"b", "c"}
-	if len(peers) != len(opts.Clustering.Peers) {
+	peers := []string{"a", "b", "c"}
+	if len(peers)-1 != len(opts.Clustering.Peers) {
 		t.Fatalf("Expected Peers to be %s, got %s", peers, opts.Clustering.Peers)
 	}
-	for i, p := range opts.Clustering.Peers {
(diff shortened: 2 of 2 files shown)


server/conf_test.go
```diff
@@ -200,15 +200,21 @@ func TestParseConfig(t *testing.T) {
 	if !opts.Clustering.Bootstrap {
 		t.Fatal("Expected Bootstrap to be true, got false")
 	}
-	peers := []string{"b", "c"}
-	if len(peers) != len(opts.Clustering.Peers) {
+	peers := []string{"a", "b", "c"}
```
kozlovic: I would revert the changes to this test. This test parses a configuration file (`test/configs/test_parse.conf`) and check that the loaded options matches what is in the file.
If you want to add a test, it would be at the end of the test `TestParseConfigureOptions`. At the end of this test, you see a case where we check node_id and peers. You could add a similar one with `a` being present on node_id and peers and check that it is then not in the array. I can add the test after merging the PR if you prefer.

server/conf.go
```diff
@@ -588,7 +588,9 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp,
 	if clusterPeers != "" {
 		sopts.Clustering.Peers = []string{}
 		for _, p := range strings.Split(clusterPeers, ",") {
-			sopts.Clustering.Peers = append(sopts.Clustering.Peers, strings.TrimSpace(p))
+			if p = strings.TrimSpace(p); p != sopts.Clustering.NodeID {
```
kozlovic: Doing the work here means that it will work in case one start the streaming server from executable, not programmatically. So I wonder if the processing should not simply be done in `RunServerWithOpts()`. But we could first go with your approach and see if it becomes an issue, move that into the run method.

primalmotion:
Ok I can revert the test, I wanted to be a good citizen :)

kozlovic:
@primalmotion And we do appreciate (even require) tests for any feature or fix. But this test here was not really testing your changes. You can either add a new test as per my comments, or I will add it afterwards. Thank you for your contribution!

kozlovic:
Thank you for your contribution!

PR #527, opened 2018-03-27
[FIXED] Small example typo in README.md [ci skip]

naphaso:


kozlovic:
Thanks! Shame on me. I added this section not long ago and had a test program to avoid this kind of issue. However, that simple part of starting a server without option, I guess I did not :-(

Issue #533, opened 2018-03-29
MySQL store sql schema file uses reserved word


camsjams:
In https://github.com/nats-io/nats-streaming-server/blob/master/mysql.db.sql

It needs a backtick around `row` to properly create the table in things like MySQL cli or workbench.

kozlovic:
Not sure why this is a problem here.. This is working fine for me. Did you actually have a problem running the NATS Streaming server against a MySQL server?

kozlovic:
Could you let me know the server version? Thanks!

camsjams:
Sure was about to add more details:
MySQL 8.0.2, (docker image - mysql:8.0.2)
Nats 0.9.0 ( docker image - nats-streaming:0.9.0)

And output from `SELECT @@GLOBAL.sql_mode;`:
`ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION
`

camsjams:
Ah but I can confirm that on MySQL 5.7 this issue does not occur.
MySQL 5.7.21, (docker image - mysql:5.7.21)
Nats 0.9.0 ( docker image - nats-streaming:0.9.0)

kozlovic:
Yes, that's the one I am using (Server version: 5.7.21 Homebrew) and did not have any issue. Regardless, I will make the appropriate changes and check if this affect Postgres. Thanks again!

PR #534, opened 2018-03-29
[FIXED] escaping of reserved word `row`

camsjams:
Fixes:
https://github.com/nats-io/nats-streaming-server/issues/533


camsjams:
If stylistically you want everything to match, I can add backticks on all column names. 

Otherwise. this should suffice and isn't too bothersome.

Issue #248, opened 2017-02-01
How can I build the server?


StarpTech:
I extract the zip and run `go build` but it throws such errors Thanks!
```
λ go build .\nats-streaming-server.go
# command-line-arguments
.\nats-streaming-server.go:115: cannot use nOpts (type *"github.com/nats-io/gnatsd/server".Options) as type *"github.com/nats-io/nats-streaming-server/vendor/github.com/nats-io/gnatsd/server".Options in argument to "github.com/nats-io/nats-streaming-server/server".ConfigureLogger
.\nats-streaming-server.go:116: cannot use nOpts (type *"github.com/nats-io/gnatsd/server".Options) as type *"github.com/nats-io/nats-streaming-server/vendor/github.com/nats-io/gnatsd/server".Options in argument to "github.com/nats-io/nats-streaming-server/server".RunServerWithOpts
```

kozlovic:
@StarpTech `go build` with vendor involved is very sensitive to the location of the source. If you already have your go environment setup, you should go into `$GOPATH/src/github.com/nats-io/nats-streaming-server` and build from there. It can't be done from another directory, and the directory must be named `nats-streaming-server`.

If you already have that directory from a previous clone, simply update it (from inside this directory):
```
# ensure you are on master
git branch
* master
# update
git pull -p
# build
go build
```

StarpTech:
@kozlovic thanks!

PR #538, opened 2018-03-30
[FIXED] SQLStore: Unable to run against MySQL 8.0.2+

### LINKED ISSUE #536: Fail to start with SQL store with MySQL server 8.0.2
@camsjams Reported this issue in #533 

MySQL 8.0.2, (docker image - mysql:8.0.2)
Nats 0.9.0 ( docker image - nats-streaming:0.9.0)
```
2018/03/29 19:39:22.709350 [INF] STREAM: Starting nats-streaming-server[test-cluster] version 0.9.0
2018/03/29 19:39:22.709553 [INF] STREAM: ServerID: dutqmYYUX3kNfne308YBqk
2018/03/29 19:39:22.709566 [INF] STREAM: Go version: go1.9.4
2018/03/29 19:39:22.721593 [INF] STREAM: Shutting down.
2018/03/29 19:39:22.721640 [FTL] STREAM: Failed to start: unable to prepare statement "INSERT INTO SubsPending (subid, row, seq) VALUES (?, ?, ?)": Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'row, seq) VALUES (?, ?, ?)' at line 1
```
As he reported, we may need to change the following: https://github.com/nats-io/nats-streaming-server/blob/4380598d34e3c8a2db9b0b1635d332639c3c6344/stores/sqlstore.go#L121-L124

Need to check the effect of Postgres driver.

kozlovic:
This is due to the use of a now reserved word. Need to be escaped
with the "`" character. This character is used only for mysql
driver, not postgres (otherwise, it would break).

Resolves #536

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -138,7 +138,7 @@ func TestMain(m *testing.M) {
 				// Create the SQL Database once, the cleanup is simply deleting
 				// content from tables (so we don't have to recreate them).
 				if err := test.CreateSQLDatabase(testSQLDriver, testSQLSourceAdmin,
-					testSQLSource, testSQLDatabaseName+n); err != nil {
+					testSQLSource+n, testSQLDatabaseName+n); err != nil {
 					fmt.Printf("Error initializing SQL Datastore: %v", err)
 					os.Exit(2)
 				}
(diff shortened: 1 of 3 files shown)


camsjams:
Thanks for quick fix! I am having trouble installing the right Go version I think, tried 1.6.2 and 1.9.1 but caught up on just building the code - which version of Go are you running locally?

Once I can get a local build I will let you know.

kozlovic:
Running go 1.10, but you could use go 1.9.4. If you can let me know what issue exactly you have, I may be able to help you. 
You may also find this useful: https://nats.io/documentation/streaming/nats-streaming-quickstart/
To get the branch with the fix, once you have cloned the repo, you would do
```
git checkout fix_536
```
Then you can build from there.

camsjams:
It was the path issue (namely where I did my `git checkout` relative to the `$GOPATH`).

Following https://nats.io/documentation/streaming/nats-streaming-quickstart/ and https://github.com/nats-io/nats-streaming-server/issues/248 helped me get the server running.

Now with go v1.10.0 and nats@nats-streaming-server#3b80e99a6698dc88f142b0ed75bb0cd27b4104de I can successfully use MySQL 8.0.2 - thanks!

kozlovic:
@camsjams Thanks for the confirmation, I am merging the PR then!

PR #551, opened 2018-04-17
[FIXED] Clustering: avoid accumulation of internal RAFT subscriptions when disconnected

### LINKED ISSUE #550: Raft stream layer accumulating subscriptions when disconnected from NATS
When the NATS transport is down, the leader from a Raft cluster will try to 'Dial' again to the other nodes that it has perceived by making a unique subscription to a subject like:

```
raft.test-cluster.$NODE_NAME.$CLUSTER.request._INBOX.$NUID

# e.g
raft.test-cluster.4tSGOGQFyl6FOd8felGDeR.test-cluster.request._INBOX.4tSGOGQFyl6FOd8felGDsZ
```

but during the reconnection these subscriptions become accumulated.

```
      "subscriptions": 107,
      "subscriptions_list": [
        "raft.test-cluster.1ZFLRZPPajd6Bj9ncuwdUJ.test-cluster.request._INBOX.1ZFLRZPPajd6Bj9ncuwrt3"
        "raft.test-cluster.1ZFLRZPPajd6Bj9ncuwdUJ.test-cluster.request._INBOX.1ZFLRZPPajd6Bj9ncuwrvT"
        "raft.test-cluster.1ZFLRZPPajd6Bj9ncuwdUJ.test-cluster.request._INBOX.1ZFLRZPPajd6Bj9ncuwrxt"
        "raft.test-cluster.1ZFLRZPPajd6Bj9ncuwdUJ.test-cluster.request._INBOX.1ZFLRZPPajd6Bj9ncuws0J"
        "raft.test-cluster.1ZFLRZPPajd6Bj9ncuwdUJ.test-cluster.request._INBOX.1ZFLRZPPajd6Bj9ncuws2j"
        "raft.test-cluster.1ZFLRZPPajd6Bj9ncuwdUJ.test-cluster.request._INBOX.1ZFLRZPPajd6Bj9ncuws59"
        "raft.test-cluster.1ZFLRZPPajd6Bj9ncuwdUJ.test-cluster.request._INBOX.1ZFLRZPPajd6Bj9ncuws7Z"
        "raft.test-cluster.1ZFLRZPPajd6Bj9ncuwdUJ.test-cluster.request._INBOX.1ZFLRZPPajd6Bj9ncuws9z"
        "raft.test-cluster.1ZFLRZPPajd6Bj9ncuwdUJ.test-cluster.request._INBOX.1ZFLRZPPajd6Bj9ncuwsCP"
         ...
```

wallyqs:
When the NATS transport is down, the leader from a Raft cluster will try to 'Dial' again to the other nodes that it has perceived by making a unique subscription to a subject like:

```
raft.test-cluster.$NODE_NAME.$CLUSTER.request._INBOX.$NUID

raft.test-cluster.4tSGOGQFyl6FOd8felGDeR.test-cluster.request._INBOX.4tSGOGQFyl6FOd8felGDsZ
```

then sending a request expecting single response, but since the transport is down these subscriptions become accumulated by the client.

In order to prevent this, a potential fix could be to try skipping Dial again in case the client is aware that it is not currently connected to NATS.

Fixes #550 

Signed-off-by: Waldemar Quevedo <wally@synadia.com>

### DIFF
--- a/server/raft_transport.go
+++ b/server/raft_transport.go
@@ -217,6 +217,10 @@ func (n *natsStreamLayer) newNATSConn(address string) *natsConn {
 // performing a handshake over NATS which establishes unique inboxes at each
 // endpoint for streaming data.
 func (n *natsStreamLayer) Dial(address raft.ServerAddress, timeout time.Duration) (net.Conn, error) {
+	if !n.conn.IsConnected() {
+		return nil, errors.New("stan: raft dial failed since not connected to NATS")
+	}
+
 	// QUESTION: The Raft NetTransport does connection pooling, which is useful
 	// for TCP sockets. The NATS transport simulates a socket using a
 	// subscription at each endpoint, but everything goes over the same NATS


server/raft_transport.go
```diff
@@ -217,6 +217,10 @@ func (n *natsStreamLayer) newNATSConn(address string) *natsConn {
 // performing a handshake over NATS which establishes unique inboxes at each
 // endpoint for streaming data.
 func (n *natsStreamLayer) Dial(address raft.ServerAddress, timeout time.Duration) (net.Conn, error) {
+	if !n.conn.IsConnected() {
+		return nil, errors.New("stan: raft dial failed since not connected to NATS")
+	}
+
```
wallyqs: Some notes around here, `n.conn.Flush()` will allow the reconnection to take up to 60 seconds, but within that time `Dial` will be called currently it seems around every second during the reconnection, so once the reconnection happens all the accumulated `Dial` calls would be unblocked and then the request happens below while it is connected again.

kozlovic: I think that we should add to the `natsStreamLayer` a field that will hold the timeout specified when creating the transport. We should then use that timeout for all Flush() calls (replace them with FlushTimeout). We also need to check the return value of the flush and unsub and return error if not nil.
Raft does "pool" the connections when possible, so it is possible that you see some of such inboxes (which correspond to raft connections if you will).

wallyqs: I agree with the fixes too, changing to `FlushTimeout` with a shorter timeout sounds better

kozlovic: The one that we don't need the actual transport timeout and check for error is in `Close()` which is when RAFT releases a connection and won't reuse it. In this case I think it could be best effort and a constant of say 500ms or 1sec is enough.

kozlovic: Do you want to have a go at the changes or do you want me to do it?

wallyqs: Looking at another issue today so feel free update, thanks Ivan!

kozlovic: I won't probably have a chance to do it today, but will update later on this week. Thanks!

PR #569, opened 2018-05-22
Update sqlstore.go

paulormart:
t as to be defined before being used in t.seq

### DIFF
--- a/stores/sqlstore.go
+++ b/stores/sqlstore.go
@@ -136,7 +136,7 @@ var sqlStmts = []string{
 	"SELECT COUNT(seq) FROM Messages WHERE id=?",                                                                                                                                     // sqlRecoverGetMessagesCount
 	"SELECT MIN(t.seq) FROM (SELECT seq FROM Messages WHERE id=? ORDER BY seq DESC LIMIT ?)t",                                                                                        // sqlRecoverGetSeqFloorForMaxMsgs
 	"SELECT COALESCE(SUM(size), 0) FROM Messages WHERE id=?",                                                                                                                         // sqlRecoverGetChannelTotalSize
-	"SELECT COALESCE(MIN(seq), 0) FROM (SELECT seq, (SELECT SUM(size) FROM Messages WHERE id=? AND seq<=t.seq) AS total FROM Messages t WHERE id=? ORDER BY seq)t2 WHERE t2.total>?", // sqlRecoverGetSeqFloorForMaxBytes
+	"SELECT COALESCE(MIN(seq), 0) FROM (SELECT seq, (SELECT SUM(size) FROM Messages t WHERE id=? AND seq<=t.seq) AS total FROM Messages WHERE id=? ORDER BY seq)t2 WHERE t2.total>?", // sqlRecoverGetSeqFloorForMaxBytes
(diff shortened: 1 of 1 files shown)


kozlovic:
@paulormart Thank you, I will look into this. As you can see, this makes tests fail and I did not have any issue with MySQL and Postgres versions I am using.
Did you have an error? If so, could you tell me what MySQL and/or Postgres versions you are using?

paulormart:
Hi @kozlovic,

The error I got:

```
panic: unable to prepare statement "SELECT COALESCE(MIN(seq), 0) FROM (SELECT seq, (SELECT SUM(size) FROM Messages WHERE id=$1 AND seq<=t.seq)
AS total FROM Messages t WHERE id=$2 ORDER BY seq)t2 WHERE t2.total>$3": pq: no data source matches prefix: t
```

It looked to me that was more sensible to define ```t``` from previous ```Messages``` table where is in use by the respective ```SELECT``` statement.

I'm using nats-streaming-server with [CockroachDB v2.0.1](https://www.cockroachlabs.com/docs/stable/build-a-go-app-with-cockroachdb.html) which is PostgresSQL compatible.

kozlovic:
Yes, but that change make the query wrong, in that it does not achieve what I was trying to do. Here is the result of a fragment on the original request with a database that contains some messages. This is running with MySQL but Postgres would be the same:
```
mysql> SELECT seq, (SELECT SUM(size) FROM Messages WHERE id=1 AND seq<=t.seq) AS total FROM Messages t WHERE id=1 ORDER BY seq;
+-----+-------+
| seq | total |
+-----+-------+
|   1 |    23 |
|   2 |   171 |
|   3 |   319 |
|   4 |   467 |
|   5 |   615 |
|   6 |   763 |
|   7 |   911 |
|   8 |  1059 |
|   9 |  1207 |
|  10 |  1355 |
|  11 |  1503 |
(...)
```
As you can see, I end-up with a running total size of all messages.
The query changed the way you have will have the total sum for each row, which is not what I want:
```
mysql> SELECT seq, (SELECT SUM(size) FROM Messages t WHERE id=1 AND seq<=t.seq) AS total FROM Messages WHERE id=1 ORDER BY seq;
+-----+-------+
| seq | total |
+-----+-------+
|   1 | 14675 |
|   2 | 14675 |
|   3 | 14675 |
|   4 | 14675 |
|   5 | 14675 |
|   6 | 14675 |
|   7 | 14675 |
|   8 | 14675 |
|   9 | 14675 |
|  10 | 14675 |
|  11 | 14675 |
(...)
```
The end goal being to delete a certain number of messages (starting from the smallest sequence number) for which their cummulative size is at least a given total.

kozlovic:
I may have found a SQL statement that would work for Postgres using `over (partition..)`, but does not work for MySQL. I may need to use different statements for this specific statement based on the driver used.

kozlovic:
@paulormart I have submitted PR #570. As mentioned there, that would fix this issue, but I believe that there will still be an issue with cockroach DB with the `SELECT FOR UPDATE` statement that we need for locking.

paulormart:
Yeah I'm aware of the ```SELECT FOR UPDATE``` hopefully cockroach db will be able to include it.
Thanks

PR #580, opened 2018-06-19
Consul support v0.10.0

xiaomi-tc:
sorry，please just ignore it.
I use nats-streaming with consul in my project and created a branch from my fork for that， and I made a mistake when I commit.


kozlovic:
I will make a more thorough review later, but at first glance here are some comments:
- In options parsing, you should simply get the consul server address, not store the consul util object in options. That object can be created in RunServerWithOpts() instead and store in server object for when you need it (for instance on shutdown).
- Tabulation seem wrong
- Please replace Chinese (?) comments with english

PR #637, opened 2018-08-10
[ADDED] Running server as a Windows Service

### LINKED ISSUE #632: Server as Windows Service
This is a duplication of the topic https://groups.google.com/forum/#!topic/natsio/jfYEfJe6bQQ
If install NATS Streaming Server as Windows Service (sc.exe create nats_streaming_server binPath = E: \ Soft \ nats-streaming-server \ nats-streaming-server.exe), the service does not start. But it would be extremely useful to have such an option (NATS Server can be installed as a service), for example, automatic start after Windows restart. In addition, it would be useful to have logging server errors in the Windows event log.
Really large companies do not use NATS Streaming Server on Windows? Run the server through the automatic launch of the bat-file at the start of the system is not a reliable solution.

kozlovic:
See README for details.

Also added option to specify a Windows Event Source name (`--syslog_name`),
which can be helpful if running more than one service on the same machine.

Resolves #632 


sharkyfm:
Maybe at the start of the service it is better to write information about the server (ServerID, Go version, nats-server version and others) into one event, rather than individual events? 21 event looks uncomfortable.

kozlovic:
I understand your concerns, however, it is not possible to do this at this stage. The logger is passed to the Store and traces about channels being recovered are traced in the Store(s) implementations. Also, NATS Streaming Server may embed the NATS Server and actually uses the NATS Server log implementation and gives that to the NATS Server embedded code to trace NATS specific events.
It would require quite a bit of refactoring to be able to do just that: coalesce the startup/recovery traces into a single trace statement.

sharkyfm:
Oh. I understood. Thanks for the work you've done. I think the functionality of the windows service will be very much not only for our company.
Ps: when is the release planned with this feature?

kozlovic:
@sharkyfm Maybe this week. Have you had a chance to try out this branch? I guess so since you reported on the logging behavior. Any other issue that you have found?

sharkyfm:
Yes, I built this branch.  The service is installed and uninstalled successfully, debug, trace and notice events are logged. Problems have not been found yet.

kozlovic:
Good to know. I will initiate the release likely today. Thanks!

PR #647, opened 2018-09-06
updating README nats.io URL ref to match new doc dir path [ci skip]

gcolliso:


gcolliso:
Merge only after new documentation goes live

PR #654, opened 2018-09-22
[IMPROVED] Reduce queue redelivery in some situations

### LINKED ISSUE #653: Don't ignore queue member ACK
When using a queue group, if a message is sent to a queue member that does not ack a message and this message gets redelivered to another member, if the first member finally acknowledges the message, and before the ackWait of the second member elapses, the server still redeliver the message.
1- server sends to member 1
2- wait (no ack from 1)
3- server sends to member 2
4- wait
5- server receives an ack from member 1
6- wait (no ack from 2)
7- server sends to member 1 again
If server was - at step 5 - removing the message as a pending from current member 2, then message would not be redelivered.

cc @soulne4ny

kozlovic:
Suppose the following example:

- send to member 1
- wait (no ack from 1)
- send to member 2
- wait
- receive an ack from member 1
- wait (no ack from 2)
- send to member 1 again

When server receives ack from member 1, although that message
is currently not assigned to this queue member, the server should
make an effort to remove this message from the ack pending list
of the member that has this message currently pending.

Resolves #653

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -3282,7 +3282,7 @@ func (s *StanServer) performAckExpirationRedelivery(sub *subState, isStartup boo
 			// We do this only after confirmation that it was successfully added
 			// as pending on the other queue subscriber.
 			if pick != sub && sent {
-				s.processAck(c, sub, m.Sequence)
+				s.processAck(c, sub, m.Sequence, false)
 			}
 		} else {
 			sub.Lock()
@@ -3313,7 +3313,7 @@ func (s *StanServer) getMsgForRedelivery(c *channel, sub *subState, seq uint64) 
 				sub.ID, seq, err)
 		}
 		// Ack it so that it does not reincarnate on restart
-		s.processAck(c, sub, seq)
+		s.processAck(c, sub, seq, false)
 		return nil
 	}
 	// The store implementation does not return a copy, we need one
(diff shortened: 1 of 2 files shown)


soulne4ny:
With release 0.11.0, it could be reproduced with this code https://gist.github.com/soulne4ny/9ef5becbbc44423b400ff12bb3fdf135

soulne4ny:
I observe that 70f729e1217525d025ae201ece12931980d548e1 behaves abnormal. After shutdown (^C) while the pair of clients are connected, following server's restart and clients' reconnect
1. It "endlessly" logs message redelivery while reconnected subscribers get no message,
2. Another client (publisher) gets `stan: publish ack timeout`.

soulne4ny:
On "endless" https://github.com/nats-io/nats-streaming-server/pull/654#issuecomment-423756616
Actually it is not endless. v0.11.0 delivers a message 3-5 times to each of two clients, while having only one in the durable queue.
Having 3 messages in the queue, one of them got delivered 8 times to client A, other – 11 times to cilent B, and the third one – 26 times to A and 30 times to B.

With 70f729e1217525d025ae201ece12931980d548e1 it looks better, but it is hard to be sure because of glitch described above.

kozlovic:
Thanks, having a look right now.

kozlovic:
@soulne4ny I have just updated the PR (force push so referenced commit id will not be valid anymore). There was a bug with locking, that is fixed. I am also now tracing only when sequence is actually found, and comes from user (I used to trace even if sequence was no longer in the sub and/or if this was an internal ack).

Note that in your test each member sleeps between 7 and 11 seconds with an AckWait of 1 second, which means that server would resend the same message many times. As explained on Slack, it means that this message would be sitting in each member internal queue several times, resulting in several "Processing ack" even after the server would stop delivering this message.

With the fix, the message will be redelivered many times until the first ack is received, but then will stop since now it does not matter if it comes from the member having the message currently assigned to. With the trace change, you will also stop seeing "Processing ack" after the message is no longer pending. The monitoring will show the 2 queue members correctly. After a server restart (if you do so after the first ack is received), you should not see any redelivery.

kozlovic:
@soulne4ny Have you had a chance to look at this? Thanks!

PR #678, opened 2018-10-24
[ADDED] serverz expose open files and max open files

caarlos0:
We are monitoring nats-streamin-service with prometheus-nats-exporter, and since nats-streaming-service doesn't have the open files count as a metric, the exporter can't export it. 

We had an issue yesterday due to "too many open files", so, this PR adds the `max_fds` and  `open_fds` metrics on OSs that have a `/proc`.

I used prometheus/procfs for that, since it is already vendored and also supports all I needed.

Hope this is something you want to add :) 

Cheers

cc/ @aricart

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -25,6 +25,7 @@ import (
 
 	gnatsd "github.com/nats-io/gnatsd/server"
 	"github.com/nats-io/nats-streaming-server/stores"
+	"github.com/prometheus/procfs"
 )
 
 // Routes for the monitoring pages
@@ -53,6 +54,8 @@ type Serverz struct {
 	Channels      int       `json:"channels"`
 	TotalMsgs     int       `json:"total_msgs"`
 	TotalBytes    uint64    `json:"total_bytes"`
+	OpenFDs       int       `json:"open_fds,omitempty"`
+	MaxFDs        int       `json:"max_fds,omitempty"`
 }
 
 // Storez describes the NATS Streaming Store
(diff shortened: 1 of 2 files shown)


server/monitor.go
```diff
@@ -53,6 +55,8 @@ type Serverz struct {
 	Channels      int       `json:"channels"`
 	TotalMsgs     int       `json:"total_msgs"`
 	TotalBytes    uint64    `json:"total_bytes"`
+	OpenFDs       int       `json:"open_fds"`
```
kozlovic: I would change to `json:"open_fds,omitempty"` and same for max_fds. This would prevent the display of those values on systems other than linux. Or would that break your prometheus monitoring tool?

server/monitor.go
```diff
@@ -187,6 +191,25 @@ func (s *StanServer) handleServerz(w http.ResponseWriter, r *http.Request) {
 	numSubs := s.numSubs
 	s.monMu.RUnlock()
 	now := time.Now()
+	pid := os.Getpid()
```
kozlovic: We may not need this.. see below.

server/monitor.go
```diff
@@ -187,6 +191,25 @@ func (s *StanServer) handleServerz(w http.ResponseWriter, r *http.Request) {
...
+	fds := 0
+	maxFDs := 0
+
+	p, err := procfs.NewProc(pid)
+	if err == nil {
```
kozlovic: Would replace this line and above with:
```
if p, err := procfs.Self(); err == nil {
```

server/monitor_test.go
```diff
@@ -233,6 +233,21 @@ func TestMonitorServerz(t *testing.T) {
...
+			t.Fatalf("Expected open files limit to be bigger than 0, got %v", sz.MaxFDs)
+		}
+	} else {
+		if sz.OpenFDs != 0 {
+			t.Fatalf("Expected 0 open files, got %v", sz.OpenFDs)
+		}
+		if sz.MaxFDs != 0 {
+			t.Fatalf("Expected open files limit to 0, got %v", sz.MaxFDs)
+		}
+	}
```
kozlovic: Assuming that it is ok to use `omitempty`, I would add a test to check that we don't remove those in the future:
```
// Not only check that values are expected to be 0, but that we don't even
// find them in the json content (omitempty)
if strings.Contains(string(body), "max_fds") || strings.Contains(string(body), "open_fds") {
    t.Fatal("open_fds and max_fds should be omitempty")
}
```

server/monitor.go
caarlos0: not a problem at all :)

server/monitor_test.go
caarlos0: done

caarlos0:
thanks for the review @kozlovic !

Changes applied as suggested

kozlovic:
@caarlos0 Thank you for the contribution!

caarlos0:
Thanks!

caarlos0:
BTW, any ETA for the next release? (no pressure, just curious :) )

kozlovic:
Did 0.11.2 not long ago and your PR is the only addition since then. No immediate plan, but may change if there are new fixes, etc..

PR #687, opened 2018-11-08
[FIXED] timestamp compare issue in GetSequenceFromTimestamp

### LINKED ISSUE #686: file store GetSequenceFromTimestamp not correct if timestamp only match the last one
I launch the nat-streaming-server with file store.
```
go run nats-streaming-server.go -st file -dir ./data
```

and the go client site
```
	scn, err := stan.Connect("test-cluster", "client")
	if err != nil {
		log.Fatal(err)
	}

	scn.Publish("subject", []byte("hello"))
	time.Sleep(300 * time.Millisecond)

	scn.Publish("subject", []byte("hello"))
	startTime := time.Now().Add(-250 * time.Millisecond)
	scn.Subscribe("subject", func(msg *stan.Msg) {
		log.Print(msg.String(), time.Unix(0, msg.Timestamp))
	}, stan.StartAtTime(startTime))

	time.Sleep(300 * time.Millisecond)
```

I expect the subscriber will get the last message, but it does not.

thlu-cobinhood:
In GetSequenceFromTimestamp, if the timestamp only match the last
message, it could ignore the last one and return ms.last + 1.

Resolve #686

### DIFF
--- a/stores/common_msg_test.go
+++ b/stores/common_msg_test.go
@@ -507,6 +507,19 @@ func TestCSGetSeqFromStartTime(t *testing.T) {
 			if seq != msgs[count-1].Sequence+1 {
 				t.Fatalf("Expected seq to be %v, got %v", msgs[count-1].Sequence+1, seq)
 			}
+
+			lastMsg := msgs[len(msgs)-1]
+			seq = msgStoreGetSequenceFromTimestamp(t, cs.Msgs, lastMsg.Timestamp)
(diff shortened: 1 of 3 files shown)


kozlovic:
@thlu-cobinhood Sorry for the late response. I would like to revisit the file store search of the correct sequence number, but I will do so after merging this PR so you can get deserved credit. Thank you for your contribution!

PR #689, opened 2018-11-10
Little typo on readme file and add a note on FT mode

ivanrodolfo:
Changes from abve to above

### DIFF
--- a/README.md
+++ b/README.md
@@ -526,7 +526,9 @@ that the RAFT state can be recovered along with the Streaming state.
 
 To minimize the single point of failure, NATS Streaming server can be run in Fault Tolerance mode. It works by having a group
 of servers with one acting as the active server (accessing the store) and handling all communication with clients, and all others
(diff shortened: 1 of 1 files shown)


kozlovic:
@ivanrodolfo Thank! When you submit a PR about a README or something that does not require tests from running you can include `[ci skip]` in the title or content of the PR and that will skip the tests (but it's fine if you omit and tests run ;-) ).

kozlovic:
Do you think you may have more fixes that you will want to add to this PR or should I merge?

ivanrodolfo:
It was my first PR here! I didn't note that will run the pipeline, by now I will not report any other change! So you can merge

kozlovic:
I wanted to add some info in FT and Clustering sections since I got several persons now mixing the two, so definitively something is missing there!
Yes, please add and I will add more later, but it's already a start and at least will alert people that they should not, even if they don't know yet why ;-)

PR #691, opened 2018-11-10
[FIXED] Channel not deleted if sub not explicitly closed

### LINKED ISSUE #690: Channel not be deleted if subscription not explicitly closed
If a connection creates a subscription and the client exits without closing the connection, the server will ultimately close it after lack of HB response. However, this does not then trigger the delete channel timer.

kozlovic:
Trigger the channel delete timer (when max inactivity is set)
when a connection is closed (either explicitly or due to lack
of HB).

Resolves #690

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -3893,7 +3893,7 @@ func (s *StanServer) removeAllNonDurableSubscribers(client *client) {
 	clientID := client.info.ID
 	client.RUnlock()
 	var (
-		storesToFlush map[string]stores.SubStore
+		storesToFlush = map[string]stores.SubStore{}
 		channels      = map[string]struct{}{}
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -3892,7 +3892,10 @@ func (s *StanServer) removeAllNonDurableSubscribers(client *client) {
 	subs := client.subs
 	clientID := client.info.ID
 	client.RUnlock()
-	var storesToFlush map[string]stores.SubStore
+	var (
+		storesToFlush map[string]stores.SubStore
```
derekcollison: If not critical performance wise just create empty storesToFlush like you do with the map below. Then you can eliminate nil checks.

PR #709, opened 2018-12-11
Update README.md

gedw99:
another silly typo...


kozlovic:
I guess I thought about `when` and `if` and settled for `if` but it became `wif` ;-)
The correction to `when` works too, so merging. Thanks!

PR #713, opened 2018-12-21
[ADDED] Encryption at rest for messages

kozlovic:
Added a new store that adds encryption of messages payload capabilities
to existing store implementations.
Server will use encryption for any selected store type if started
with the Encrypt option set to true and the EncryptionKey string.
The environment variable NATS_STREAMING_ENCRYPTION_KEY will be used
if set. If not, will use the EncryptionKey option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ NATS Streaming provides the following high-level feature set.
             * [Queue Group](#queue-group)
             * [Redelivery](#redelivery)
     * [Store Interface](#store-interface)
+    * [Store Encryption](#store-encryption)
     * [Clustering](#clustering)
         * [Supported Stores](#supported-stores)
         * [Clustering Configuration](#clustering-configuration)
@@ -316,6 +317,37 @@ If you wish to contribute to a new store type, your implementation must include 
 The memory and the provided file store implementations both use a generic store implementation to avoid code duplication.
 When writing your own store implementation, you can do the same for APIs that don't need to do more than what the generic implementation provides.
 You can check [MemStore](https://github.com/nats-io/nats-streaming-server/blob/master/stores/memstore.go) and [FileStore](https://github.com/nats-io/nats-streaming-server/blob/master/stores/filestore.go) implementations for more details.
+
+## Store Encryption
+
+The server can be configured to encrypt messages payload when storing them, providing encryption at rest.
+This can be done from the command line or from the configuration file. Check `encrypt` and `encryption_key`
+in the [Configuring](#configuring) section.
+
+It is recommended to provide the encryption key through the environment variable `NATS_STREAMING_ENCRYPTION_KEY`
+instead of `encryption_key`. If encryption is enabled and `NATS_STREAMING_ENCRYPTION_KEY` is found, this
+will take precedence over `encryption_key` value.
+
+You can pass this from the command line this way:
+```
+$ env NATS_STREAMING_ENCRYPTION_KEY="mykey" nats-streaming-server -store file -dir datastore -encrypt
+```
+
+Note that only message payload is encrypted, all other data stored by NATS Streaming server is not.
+
+When running in clustering mode (see below), the server uses RAFT, which uses its own log files.
+Those will be encrypted too.
+
+Starting a server with `encrypt` against a datastore that was not encrypted will result in failures
+when it comes to decrypt a message, which may not happen right during the startup process. Instead,
+it will happen when attempting to deliver messages to consumers.
+
+The same behavior will occur if the encryption key is not the one that was used to encrypt the data.
+
+Performance considerations: As expected, encryption is likely to decrease performance, but by how much is hard
+to define. In some performance tests on a MacbookPro 2.8 GHz Intel Core i7 with SSD, we have
+observed as little as 1% decrease to more than 30%. In addition to CPU cycles required for encryption,
+the encrypted payload is bigger, which result in more data being stored or read.
 
 ## Clustering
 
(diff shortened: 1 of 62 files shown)


README.md
```diff
@@ -317,6 +318,37 @@ The memory and the provided file store implementations both use a generic store
 When writing your own store implementation, you can do the same for APIs that don't need to do more than what the generic implementation provides.
 You can check [MemStore](https://github.com/nats-io/nats-streaming-server/blob/master/stores/memstore.go) and [FileStore](https://github.com/nats-io/nats-streaming-server/blob/master/stores/filestore.go) implementations for more details.
 
+## Store Encryption
+
+The server can be configured to encrypt messages payload when storing them, providing encryption at rest.
```
derekcollison: encrypt a message's

README.md
```diff
@@ -317,6 +318,37 @@ The memory and the provided file store implementations both use a generic store
...
+$ env NATS_STREAMING_ENCRYPTION_KEY="mykey" nats-streaming-server -store file -dir datastore -encrypt
+```
+
+Note that only message payload is encrypted, all other data stored by NATS Streaming server is not.
+
+When running in clustering mode (see below), the server uses RAFT, which uses its own log files.
+Those will be encrypted too.
+
+Starting a server with `encrypt` against a datastore that was not encrypted will result in failures
+when it comes to decrypt a message, which may not happen right during the startup process. Instead,
```
derekcollison: which may not happen immediately upon startup.

README.md
```diff
@@ -317,6 +318,37 @@ The memory and the provided file store implementations both use a generic store
...
+it will happen when attempting to deliver messages to consumers.
+
+The same behavior will occur if the encryption key is not the one that was used to encrypt the data.
```
derekcollison: Should we have tooling to move a store from plaintext to encrypted?

README.md
```diff
@@ -317,6 +318,37 @@ The memory and the provided file store implementations both use a generic store
...
+
+Performance considerations: As expected, encryption is likely to decrease performance, but by how much is hard
+to define. In some performance tests on a MacbookPro 2.8 GHz Intel Core i7 with SSD, we have
+observed as little as 1% decrease to more than 30%. In addition to CPU cycles required for encryption,
+the encrypted payload is bigger, which result in more data being stored or read.
```
derekcollison: Why is it bigger?

derekcollison: Maybe we should also compress it like I do now for the usage server.

README.md
```diff
@@ -1494,6 +1528,8 @@ In general the configuration parameters are the same as the command line argumen
 | ft_group | In Fault Tolerance mode, you can start a group of streaming servers with only one server being active while others are running in standby mode. This is the name of this FT group | String | `ft_group: "my_ft_group"` |
 | partitioning | If set to true, a list of channels must be defined in store_limits/channels section. This section then serves two purposes, overriding limits for a given channel or adding it to the partition | `true` or `false` | `partitioning: true` |
 | cluster | Cluster Configuration | Map: `cluster: { ... }` | **See details below** |
+| encrypt | Specify if server should encrypt messages (only the payload) when storing them | `true` or `false` | `encrypt: true` |
```
derekcollison: We don't talk anywhere about key rotation.

server/raft_log.go
```diff
@@ -63,6 +80,35 @@ func newRaftLog(log logger.Logger, fileName string, sync bool, trailingLogs int)
...
+		defer func() {
+			for i := 0; i < len(encryptionKey); i++ {
+				encryptionKey[i] = 'x'
+			}
+		}()
+
+		h := sha256.New()
+		h.Write(key)
+		keyHash := h.Sum(nil)
+		gcm, err := chacha20poly1305.New(keyHash)
```
derekcollison: Should this be settable longer term?

README.md
kozlovic: Going to be painful.. What I could do (and we should decide before accepting PR) is to had a few bytes at the beginning of an encrypted payload (before we copy the nonce) to indicate if the data is encrypted or not. So a store with encryption would look at the beginning of the data and if it does not find the marker return the payload as-is (assuming not encrypted). If the marker is present, decrypt. Obviously, if non-encrypted data starts with this marker, then this would result in decrypt failure.

README.md
kozlovic: Crypto overhead.

README.md
kozlovic: Key is not rotated. The nonce is generated after a random number of time it was used for encryption. What do you have in mind?

server/raft_log.go
kozlovic: Settable how? chacha20poly1305.New() requires a 256 bit key, which is what the few lines above provide.

derekcollison: Would default to AES, faster on Intels and most server boxes. ChaCha good for ARM devices.

README.md
derekcollison: We can solve that later..

server/raft_log.go
derekcollison: Add one byte indicating algorithm to use..

kozlovic:
@derekcollison I think I have addressed the issues we discussed.

stores/cryptostore.go
```diff
@@ -0,0 +1,329 @@
...
+			return 0, nil, ErrCipherNotSupported
+		}
+	} else {
+		// Otherwise default to AES on intel (there is hardware
+		// acceleration for that) and chacha20poly1305 on ARM
+		// (the two arch'es that we build docker images for).
+		if runtime.GOARCH == "amd64" || runtime.GOARCH == "386" {
+			code = CryptoCodeAES
+		} else {
+			fmt.Printf("@@IK: selected CHACHA\n")
```
wallyqs: Think maybe leftover debug comment here at line 122: `fmt.Printf("@@IK: selected CHACHA\n")`

derekcollison: Good catch!

PR #734, opened 2019-01-29
docs: Cluster Bootstrapping

akoenig:
I tried the auto configuration feature and encountered that the `-nats_server nats://localhost:4222` argument shouldn't be defined when starting the leader node. Is that correct? 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,7 +539,7 @@ Here is an example of starting three servers in a cluster by starting one as
 the seed and letting the others automatically join:
 
 ```
-nats-streaming-server -store file -dir store-a -clustered -cluster_bootstrap -nats_server nats://localhost:4222
+nats-streaming-server -store file -dir store-a -clustered -cluster_bootstrap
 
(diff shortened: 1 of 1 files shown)


kozlovic:
This example in the doc simply assumes that you are running a standalone NATS Server to which all three servers connect to. At the beginning of the clustering configuration section, there is this comment:
```
Here is an example of starting three servers in a cluster. For this example, we run a separate NATS server which the Streaming servers connect to.
```
which is used many times in the rest of the doc. If you do not want to use a standalone NATS Server then you would need to define routes to each other and have NATS Streaming server embeds the NATS Server.
If you remove `-nats_server` from the first server, you simply make this Streaming server start with its embedded NATS Server, and the two other that you start (with `-nats_server`) would then not embed a NATS Server but connect to the NATS Server embedded by the first server. Although it would work, you would need to understand what that means, and this is probably not a good approach. I would recommend re-reading the very first section of the README that explains relation between NATS and NATS Streaming: https://github.com/nats-io/nats-streaming-server#relation-to-nats

PR #738, opened 2019-02-05
[FIXED] Delayed msg processing when getting invalid pub requests

kozlovic:
When a published message is processed by the server and the
associated clientID is not found, in partitions mode, we need
to wait to see if the processing of the connect is going to
happen soon. Delay here is delaying all other inbound messages.
We don't have a choice here, but reducing the wait time and
also remembering the failed clientIDs so that next invalid
message is processed fast (map lookup).

Also removed this check from clustered mode. It used to be required
when clustering was using a RAFT group per channel and one for
meta data, but this is no longer the case (single RAFT group).

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -21,12 +21,18 @@ import (
 	"github.com/nats-io/nats-streaming-server/stores"
 )
 
+const (
+	maxKnownInvalidConns   = 256
+	pruneKnownInvalidConns = 32
+)
+
 // This is a proxy to the store interface.
 type clientStore struct {
 	sync.RWMutex
 	clients        map[string]*client
 	connIDs        map[string]*client
 	waitOnRegister map[string]chan struct{}
+	knownInvalid   map[string]struct{}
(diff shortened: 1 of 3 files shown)


server/client.go
```diff
@@ -145,11 +159,41 @@ func (cs *clientStore) isValidWithTimeout(ID string, connID []byte, timeout time
...
 		cs.Lock()
 		delete(cs.waitOnRegister, ID)
+		cs.addToKnownInvalid(ID, connID)
 		cs.Unlock()
 		return false
 	}
 }
 
+func (cs *clientStore) addToKnownInvalid(ID string, connID []byte) {
+	cID := string(connID)
```
derekcollison: Since you do this twice maybe make it a small func, would simplify knownToBeInvalid as well.

kozlovic: I actually moved to using clientID+":"+connID as the key of the map. It simplifies the code I think. If you could have a second look. Thanks!

PR #756, opened 2019-02-28
[ADDED] GitCommit to NATS Streaming Server

### LINKED ISSUE #755: can't set `version.GITCOMMIT`
From the Dockerfile, `GITCOMMIT` is set through `ldflags`, which is supposed to locate in the package `github.com/nats-io/nats-streaming-server/version`. But there is no such package and no corresponding statements found to print this info.

kozlovic:
Also adds the proper git commit for the embedded NATS Server.

Resolves #755

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,11 +5,16 @@ MAINTAINER Ivan Kozlovic <ivan@synadia.com>
 COPY . /go/src/github.com/nats-io/nats-streaming-server
 WORKDIR /go/src/github.com/nats-io/nats-streaming-server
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,15 +1,20 @@
-FROM golang:1.11.5
+FROM golang:1.12
```
kozlovic: Maybe not move docker image to 1.12 yet (at least in this PR). Since we may do a 0.12.2 soon, we may want to stay at same go release of 1.11.5 compared to streaming 0.12.0.

PR #767, opened 2019-03-11
Allow custom nats client options server configuration

danielsdeleo:
## Summary

Allows custom NATS client options to be set on the internal NATS client connection.

## Rationale

I am integrating NATS streaming into Chef's Automate product. Automate internally has a Microservices architecture with connections between services secured by mutual TLS cert verification. Automate is usually deployed by our customers on their own infrastructure so we have elected not to rely on using hostnames in our certificates' CN fields; instead we use the internal service names (i.e., something like `event-service` instead of `host.example.com`). In go it's fairly straightforward to adapt to this by setting the `ServerName` field of the `tls.Config` struct when it's exposed, as the NATS go client does; however NATS streaming does not currently expose the full set of NATS client options.

## Tests?

Yep! Happy to add tests, I just wanted to ensure I'm taking the preferred approach for implementing this functionality first.


kozlovic:
This looks good to me. I may wait before merging that to master because we are about to do a 0.12.2 release and I would not want to add a new option for that update release, but will do it just after.

PR #776, opened 2019-03-15
Release v0.12.2

kozlovic:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.11.5
+FROM golang:1.11.6
 
 MAINTAINER Ivan Kozlovic <ivan@synadia.com>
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.11.5
+FROM golang:1.11.6
```
kozlovic: Actually, looks like docker image 1.11.6 not yet available. So if doing release today, will have to keep 1.11.5

kozlovic: Will keep checking on avail.. may be coming today?

derekcollison: idk

kozlovic: avail now, so merging.

PR #779, opened 2019-03-20
[FIXED] FileStore: possible panic when removing old messages

### LINKED ISSUE #778: v0.11.2 Panic in file store removeFirstMsg
`panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x84ddf4]

goroutine 33 [running]:
github.com/nats-io/nats-streaming-server/stores.(*FileMsgStore).removeFirstMsg(0xc00001e4e0, 0x0, 0xa70000)
        E:/GoProjs/src/github.com/nats-io/nats-streaming-server/stores/filestore.go:2928 +0x64
github.com/nats-io/nats-streaming-server/stores.(*FileMsgStore).enforceLimits(0xc00001e4e0, 0xf0001, 0xc000ce47e0, 0xc001fd6101)
        E:/GoProjs/src/github.com/nats-io/nats-streaming-server/stores/filestore.go:2870 +0x95
github.com/nats-io/nats-streaming-server/stores.(*FileMsgStore).Store(0xc00001e4e0, 0xc000ce47e0, 0x0, 0x0, 0x0)
        E:/GoProjs/src/github.com/nats-io/nats-streaming-server/stores/filestore.go:2713 +0xaf2
github.com/nats-io/nats-streaming-server/server.(*StanServer).ioLoop.func1(0xc000278000, 0x1, 0x400)
        E:/GoProjs/src/github.com/nats-io/nats-streaming-server/server/server.go:3674 +0x23d
github.com/nats-io/nats-streaming-server/server.(*StanServer).ioLoop(0xc0001e4000, 0xc00012e2c0)
        E:/GoProjs/src/github.com/nats-io/nats-streaming-server/server/server.go:3762 +0x3ba
created by github.com/nats-io/nats-streaming-server/server.(*StanServer).startIOLoop
        E:/GoProjs/src/github.com/nats-io/nats-streaming-server/server/server.go:3603 +0xcc`

we start with command:
`nohup ./nats-streaming-server --store FILE --dir /tmp/nats --user foo --pass bar -m 8222 &`
it's ok after reboot the server

kozlovic:
When removing the first message in a log, there is a possibility
that an I/O error occurs, which then translated to an object
being nil and being dereferenced, which would cause a panic.

Resolves #778

### DIFF
--- a/stores/filestore.go
+++ b/stores/filestore.go
@@ -2827,7 +2827,16 @@ func (ms *FileMsgStore) expireMsgs(now, maxAge int64) int64 {
 				}
 			}
 			if slice != nil {
-				m = ms.getMsgIndex(slice, ms.first)
+				var err error
+				if m, err = ms.getMsgIndex(slice, ms.first); err != nil {
+					ms.log.Errorf("Error during expiration: %v", err)
+					if slice != nil {
+						ms.unlockIndexFile(slice)
+					}
+					// Try again in 5 secs.
+					ms.expiration = now + int64(5*time.Second)
+					return ms.expiration
+				}
 			}
 		}
 		if m == nil {
@@ -2867,7 +2876,13 @@ func (ms *FileMsgStore) enforceLimits(reportHitLimit, lockFile bool) error {
 
 		// Remove first message from first slice, potentially removing
 		// the slice, etc...
-		ms.removeFirstMsg(nil, lockFile)
+		if err := ms.removeFirstMsg(nil, lockFile); err != nil {
+			// We are not going to fail the publish, just report
+			// the error removing the first message.
+			// TODO: Is this the right thing to do?
+			ms.log.Errorf("Unable to remove first message: %v", err)
+			return nil
+		}
 		if reportHitLimit && !ms.hitLimit {
 			ms.hitLimit = true
 			ms.log.Warnf(droppingMsgsFmt, ms.subject, ms.totalCount, ms.limits.MaxMsgs,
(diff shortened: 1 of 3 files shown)


stores/filestore.go
```diff
@@ -2867,7 +2877,13 @@ func (ms *FileMsgStore) enforceLimits(reportHitLimit, lockFile bool) error {
 
 		// Remove first message from first slice, potentially removing
 		// the slice, etc...
-		ms.removeFirstMsg(nil, lockFile)
+		if err := ms.removeFirstMsg(nil, lockFile); err != nil {
+			//We are not going to fail the publish, just report
```
derekcollison: spc

stores/filestore.go
```diff
@@ -2827,7 +2827,17 @@ func (ms *FileMsgStore) expireMsgs(now, maxAge int64) int64 {
 				}
 			}
 			if slice != nil {
-				m = ms.getMsgIndex(slice, ms.first)
+				var err error
+				m, err = ms.getMsgIndex(slice, ms.first)
```
derekcollison: I think could be one line since not assigning. `if m, err = ..; err != nil {`

stores/filestore.go
```diff
@@ -3254,10 +3278,18 @@ func (ms *FileMsgStore) GetSequenceFromTimestamp(timestamp int64) (uint64, error
 			return 0, err
 		}
 		seq := slice.firstSeq
-		firstMsgInSlice := ms.getMsgIndex(slice, seq)
+		firstMsgInSlice, err := ms.getMsgIndex(slice, seq)
+		if err != nil {
```
derekcollison: Same as above.

kozlovic: This one no, because I need firstMsgInSlice after that... unless I declare them both with `var` before that...

stores/filestore.go
kozlovic: ok

stores/filestore.go
kozlovic: ok

stores/filestore.go
kozlovic: .. or I use `else` statements (since the local declaration would be visible there)

derekcollison: Agree with you, since its declaration.

kozlovic: I decided to declare in the `if` actually, and since it is visible in `else` and not used after that.

kozlovic:
(did a rebase - to remove Go 1.10 from travis, and squashed commits into 1).

derekcollison:
LGTM

PR #786, opened 2019-03-26
fix: mysql max column and key size

ekoeryanto:


kozlovic:
@ekoeryanto Would it be possible for you to point to documentation as to why this needs to be 191? From the doc, I see that it looks like max should be set to 255 (and not 256) assuming utf8 and all characters 3 bytes, since 255*3=765 which is below the 767 limit.

ekoeryanto:
>  ...assuming utf8 and all characters 3  bytes, since 255*3=765 which is below the 767 limit.

Yes, I read that somewhere in the doc, but when I tried to execute all of the _queries_ with collation `utf8_unicode_ci` or `utf8mb4_unicode_ci` both of `Clients` and `Channels` table throw errors

kozlovic:
Humm... what version of MySQL are you using? I did not get such reports from other users. It may be something specific to your environment. At this point I would not make the change to 191 in the repo, but obviously if it works for you use that value. As mentioned in the README, this is just to help users kick start the deployment, DBAs are encouraged to tune for optimal performance.
So if anything, I would agree in lowering to 255 (from 256), but not to 191 without some doc reference that explains that this is the max primary key length for a VARCHAR.
I hope you understand.

ekoeryanto:
> At this point I would not make the change to 191 in the repo

That's pair enough.

Ah I see, I tried on mysql 5.7, it just work, but not with mariadb 10.x.

kozlovic:
@ekoeryanto Yes, so each DB may have its specific requirements. So end-user may need to fine tune the script.

PR #810, opened 2019-04-24
[FIXED] Clustering: erroneous "too many subscriptions per channel"

### LINKED ISSUE #809: Clustering: Version 0.14.0 produces erroneous "too many subscriptions per channel"
Version 0.14.0 has introduced a defect that would cause "too many subscriptions per channel" errors even though the actual number of subscriptions is below the limit.

cc @lazyshot

kozlovic:
This was introduced by #802. The server was replacing the sub's ID
after the call to CreateSub(), but that meant that DeleteSub() in
the underlying sub store would not find that subscription, which
will then cause it to report too many subscriptions after a while
because the removed subscriptions were not removed from their
internal map.
The approach in this PR is to make RaftSubStore completely bypass
the underlying SubStore. It should work with a mix of servers
of different versions.

Resolves #809

### DIFF
--- a/stores/raftstore.go
+++ b/stores/raftstore.go
@@ -16,6 +16,7 @@ package stores
 import (
 	"sync"
 
+	"github.com/nats-io/nats-streaming-server/logger"
 	"github.com/nats-io/nats-streaming-server/spb"
 )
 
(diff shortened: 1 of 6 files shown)


stores/raftstore.go
```diff
@@ -93,6 +116,45 @@ func (s *RaftStore) DeleteClient(clientID string) error {
 // RaftSubStore methods
 ////////////////////////////////////////////////////////////////////////////
 
+// CreateSub implements the SubStore interface
+func (ss *RaftSubStore) CreateSub(sub *spb.SubState) error {
+	gss := &ss.genericSubStore
+
+	gss.Lock()
+	defer gss.Unlock()
```
derekcollison: Does perf matter here?

kozlovic: Not that many subs can be created per second, so not really. Plus, although as you know I am in favor of using `defer` sparingly, when you start to have multiple returns in the function, it starts to make sense.

derekcollison: Agree

PR #827, opened 2019-05-06
[FIXED] Incorrect number of subscriptions in /streaming/serverz

### LINKED ISSUE #819: Incorrect number of subscriptions in `streaming/serverz` when durables are resumed
The reported number of subscriptions increases every time a durable (queue or not) is restarted.

kozlovic:
Replaces the use of a variable that was supposed to be updated
on subscription create/remove. It was started to be a bit brittle.

Provide a function that computes the number of subscriptions
(including offline durables/queue durables).
This is a bit more costly than using the variable but will be
less prone to issues.

Resolves #819

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -605,10 +605,6 @@ type StanServer struct {
 	// Store
 	store stores.Store
 
-	// Monitoring
-	monMu   sync.RWMutex
-	numSubs int
-
 	// IO Channel
 	ioChannel     chan *ioPendingMsg
 	ioChannelQuit chan struct{}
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -742,6 +738,36 @@ type subSentAndAck struct {
...
+			// If this is a durable queue subscription and all members
+			// are offline, qsub.shadow will be not nil. Report this one.
+			if qsub.shadow != nil {
+				total++
+			}
+			qsub.RUnlock()
+		}
+		c.ss.RUnlock()
+	}
+	s.channels.RUnlock()
```
derekcollison: With cascading three tiers of locks, do we know we do not violate order somewhere else?

kozlovic: We do not. This is a well known order: the channel store is the main, then under this one we can take the channel subStore lock. Then we possibly go over queue states.
We do this on message delivery, monitor gathering of subs, etc..

derekcollison: ok

PR #828, opened 2019-05-09
[FIXED] SQLStore: panic when updating subscriptions pending state

### LINKED ISSUE #741: panic: Unable to flush sub store: pq: duplicate key value violates unique constraint "pk_msgpendingkey"
Got this error yesterday and today, and it's fixed after deleting the table contents by myself.

```
panic: Unable to flush sub store: pq: duplicate key value violates unique constraint "pk_msgpendingkey"
goroutine 22 [running]:
github.com/nats-io/nats-streaming-server/server.(*StanServer).ioLoop(0xc0001e6000, 0xc000146630)
    /home/admin/nats/src/github.com/nats-io/nats-streaming-server/server/server.go:3782 +0xb4d
created by github.com/nats-io/nats-streaming-server/server.(*StanServer).startIOLoop
    /home/admin/nats/src/github.com/nats-io/nats-streaming-server/server/server.go:3611 +0xcc
```

Version :
version | 1.3.0
proto | 1
go | go1.11.4
Postgres : 10 on AWS RDS

kozlovic:
Issue was due to the fact that we were not keeping track of the
maximum row number, so if a subscription that are pending state
saved in row 1 and 3 and the last recovered subscription had
state in row 2, and then the next update was for the first subscription
the SQL statement would have tried to add a row with ID 3 which
already existed.

Resolves #741

### DIFF
--- a/stores/sqlstore.go
+++ b/stores/sqlstore.go
@@ -104,10 +104,10 @@ var sqlStmts = []string{
 	"INSERT INTO Channels (id, name, maxmsgs, maxbytes, maxage) VALUES (?, ?, ?, ?, ?)",                          // sqlAddChannel
 	"INSERT INTO Messages VALUES (?, ?, ?, ?, ?)",                                                                // sqlStoreMsg
 	"SELECT timestamp, data FROM Messages WHERE id=? AND seq=?",                                                  // sqlLookupMsg
-	"SELECT seq FROM Messages WHERE id=? AND timestamp>=? LIMIT 1",                                               // sqlGetSequenceFromTimestamp
+	"SELECT seq FROM Messages WHERE id=? AND timestamp>=? ORDER BY seq LIMIT 1",                                  // sqlGetSequenceFromTimestamp
 	"UPDATE Channels SET maxseq=? WHERE id=?",                                                                    // sqlUpdateChannelMaxSeq
 	"SELECT COUNT(seq), COALESCE(MAX(seq), 0), COALESCE(SUM(size), 0) FROM Messages WHERE id=? AND timestamp<=?", // sqlGetExpiredMessages
-	"SELECT timestamp FROM Messages WHERE id=? AND seq>=? LIMIT 1",                                               // sqlGetFirstMsgTimestamp
+	"SELECT timestamp FROM Messages WHERE id=? AND seq>=? ORDER BY seq LIMIT 1",                                  // sqlGetFirstMsgTimestamp
 	"DELETE FROM Messages WHERE id=? AND seq<=?",                                                                 // sqlDeletedMsgsWithSeqLowerThan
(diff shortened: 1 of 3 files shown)


stores/sqlstore.go
```diff
@@ -104,10 +104,10 @@ var sqlStmts = []string{
 	"INSERT INTO Channels (id, name, maxmsgs, maxbytes, maxage) VALUES (?, ?, ?, ?, ?)",                          // sqlAddChannel
 	"INSERT INTO Messages VALUES (?, ?, ?, ?, ?)",                                                                // sqlStoreMsg
 	"SELECT timestamp, data FROM Messages WHERE id=? AND seq=?",                                                  // sqlLookupMsg
-	"SELECT seq FROM Messages WHERE id=? AND timestamp>=? LIMIT 1",                                               // sqlGetSequenceFromTimestamp
+	"SELECT seq FROM Messages WHERE id=? AND timestamp>=? ORDER BY seq LIMIT 1",                                  // sqlGetSequenceFromTimestamp
```
derekcollison: Do we have best practices or whatever to setup correct indexes on the DB? Does this affect perf at all?

derekcollison: If you say limit 1 why do you need "order by"?

kozlovic: I forgot to add a comment that these changes are actually not required for the fix, it is just that when I investigated, I read that select order is not guaranteed unless ORDERY BY is specified: the doc (https://www.postgresql.org/docs/8.1/queries-order.html) says: `After a query has produced an output table (after the select list has been processed) it can optionally be sorted. If sorting is not chosen, the rows will be returned in an unspecified order`.
We have indexes and seq is the primary key so that should not impact perf, and my assumption was that I did not need ORDER BY. But if for some reason the select is not producing the results in the order I think is implied, then it could be wrong. Imagine the following list of messages:
```
seq: 1 - timestamp: 100
seq: 2 - timestamp: 200
seq: 3 - timestamp: 300
seq: 4 - timestamp: 400
```
With the above order, you expect `SELECT seq FROM Messages WHERE timestamp>=300 LIMIT 1` to return `seq==3`. But since doc says that order is not guaranteed unless ORDER BY is specified, it should then be possible that the above list for SELECT appears as:
```
seq: 1 - timestamp: 100
seq: 4 - timestamp: 400
seq: 2 - timestamp: 200
seq: 3 - timestamp: 300
```
and in this case the query would return `seq==4`. I cannot reproduce this, but wonder if this could happen. Writing the statement as `SELECT seq FROM Messages WHERE timestamp>=300 ORDER BY seq LIMIT 1` would give the expected result because seq and timestamp are proxies (we make sure that messages are stored with increasing seq/timestamp combo), but for two messages with the same timestamp, we would want the message with smallest sequence.

derekcollison: In this case you are trying to get the starting sequence for a replay?

kozlovic: correct

PR #840, opened 2019-05-18
[FIXED] Clustering: channel first/last sequence may fall to zero

### LINKED ISSUE #833: Clustering: still seeing cases where store first/last falls to 0
If a leader has a channel for which all messages have expired and no new message is sent on that channel, and a snapshot is taken and there is no raft log after that for this channel, a node that is restarted without state would restore from the snapshot but since all messages are expired, it would not persist anything in the channel's store.
If later this node becomes leader, and still no message is produced on this channel, and a snapshot is taken, it would persist first/last == 0 in the raft snapshot, which means that if yet another node is restarted without state, it would then get that there is no message in the channel and does not know that first sequence should be something else. So if/when that last node becomes leader later on, then it would start storing messages at sequence 1 for this channel, and all existing subscriptions would not receive messages until the first sequence in the channel reaches their last_sent value.

kozlovic:
This could happen if the leader took a snapshot while messages were
not yet expired, then a node is started without state and tries
to restore from this snapshot. If the messages have expired by then,
no message would be stored. If that node later did a snapshot itself,
it would persist in it the first/last being zero. If no message are
published and this node becomes leader, it would start storing
messages at the wrong sequence and would also send the bad snapshot
to other nodes.

Resolves #833

### DIFF
--- a/server/clustering.go
+++ b/server/clustering.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/hashicorp/raft"
@@ -446,6 +447,32 @@ func (s *StanServer) getClusteringPeerAddr(raftName, nodeID string) string {
 	return fmt.Sprintf("%s.%s.%s", s.opts.ID, nodeID, raftName)
 }
 
+// Returns the message store first and last sequence.
+// When in clustered mode, if the first and last are 0, returns the value of
+// the last sequence that we possibly got from the last snapshot. If a node
+// restores a snapshot that let's say has first=1 and last=100, but when it
+// tries to get these messages from the leader, the leader does not send them
+// back because they have all expired, the node will not store anything.
+// If we just rely on store's first/last, this node would use and report 0
+// for channel's first and last while when all messages have expired, it should
+// be last+1/last.
+func (s *StanServer) getChannelFirstAndlLastSeq(c *channel) (uint64, uint64, error) {
+	first, last, err := c.store.Msgs.FirstAndLastSequence()
+	if !s.isClustered {
+		return first, last, err
+	}
+	if err != nil {
+		return 0, 0, err
+	}
+	if first == 0 && last == 0 {
+		if fseq := atomic.LoadUint64(&c.firstSeq); fseq != 0 {
+			first = fseq
+			last = fseq - 1
+		}
+	}
+	return first, last, nil
+}
+
 // Apply log is invoked once a log entry is committed.
 // It returns a value which will be made available in the
 // ApplyFuture returned by Raft.Apply method if that
@@ -456,13 +483,16 @@ func (r *raftFSM) Apply(l *raft.Log) interface{} {
 	if err := op.Unmarshal(l.Data); err != nil {
 		panic(err)
 	}
+	// We don't want snapshot Persist() and Apply() to execute concurrently,
+	// so use common lock.
+	r.Lock()
+	defer r.Unlock()
 	switch op.OpType {
 	case spb.RaftOperation_Publish:
 		// Message replication.
 		var (
-			c       *channel
-			err     error
-			lastSeq uint64
+			c   *channel
+			err error
 		)
 		for _, msg := range op.PublishBatch.Messages {
 			// This is a batch for a given channel, so lookup channel once.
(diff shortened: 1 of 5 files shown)


server/clustering.go
```diff
@@ -477,18 +507,25 @@ func (r *raftFSM) Apply(l *raft.Log) interface{} {
...
-			if err == nil && lastSeq < msg.Sequence-1 {
-				err = s.raft.fsm.restoreMsgsFromSnapshot(c, lastSeq+1, msg.Sequence-1)
 			}
 			if err == nil {
 				_, err = c.store.Msgs.Store(msg)
 			}
 			if err != nil {
-				return fmt.Errorf("failed to store replicated message %d on channel %s: %v",
-					msg.Sequence, msg.Subject, err)
+				panic(fmt.Errorf("failed to store replicated message %d on channel %s: %v",
```
derekcollison: ok so this I am sure is on purpose since there is nothing we can do here. Just checking.

server/snapshot.go
```diff
@@ -326,31 +337,35 @@ func (r *raftFSM) restoreChannelsFromSnapshot(serverSnap *spb.RaftSnapshot, inNe
...
+		// If we skip it (like we use to), then it is possible that this
+		// node misses some messages from the snapshot and will then start
+		// to apply remaining logs. It could even then become leader in
+		// the process with missing messages. So if we need to restore some
+		// messages and fail, keep trying until a leader is avail and
+		// able to serve this node.
+		for {
+			if err := r.restoreMsgsFromSnapshot(c, sc.First, sc.Last, false); err != nil {
+				s.log.Errorf("unable to restore messages, can't start until leader is available")
+				time.Sleep(time.Second)
```
derekcollison: Is a whole second required here?

server/clustering.go
kozlovic: Yes, it was a panic before, changed it (in master - no public release) to a return error, but it has to be a panic. RAFT has committed a log and applies it to the FSM (the streaming server code). If we fail to store in streaming store, we should bail and not continue (RAFT would simply move to next raft log). You would not even see the error since nothing is waiting/reporting on this apply future.

server/snapshot.go
kozlovic: Yes.. the restore failure is either because the leader is gone, or if it is a storage failure, it is likely not going to recover anyway. If it is a failure because leader is not running/gone, a new leader election takes some time, so no reason to rush this.

kozlovic: Not to say that those are the only 2 reasons, there could be NATS communication issues (like sending the request fails due to disconnect and need to wait for reconnect), etc..

derekcollison: Do we ever exit this or can a server look stuck sans the log messages? Should we bail after some point?

kozlovic: No, this is an infinite loop. We could exit after few attempts but the point is that server will always be stuck at this step until a leader is able to service it. I can make it exit/panic after a certain number of attempts if you think it is better.

kozlovic: And we want this server to either complete this step without errors or stay there (or exit, but the step can't be skipped, so will get back there on restart).

derekcollison: Any management system would be waiting for it to exit to take action IMO. And if it was a health probe it would need to be able to signal to kill the server. I think we try for say 15-3o seconds then exit(-1).

kozlovic:
@derekcollison I have pushed forced because I did a rebase from master to take the raft shutdown changes. But I kept a separate commit to ease the review. I can squash prior to merge. Commit to review would be this [one](https://github.com/nats-io/nats-streaming-server/pull/840/commits/fbc4114e0dece0a95966e346b65a24fe713d0ff7)

server/snapshot.go
```diff
@@ -234,6 +246,27 @@ func (r *raftFSM) Restore(snapshot io.ReadCloser) (retErr error) {
 	r.Lock()
 	defer r.Unlock()
 
+	shouldSnapshot := false
+	defer func() {
+		if retErr == nil && shouldSnapshot {
```
derekcollison: Doesn't this snapshot retErr here when defining the func? Meaning if you set it later it won't matter? May be remembering incorrectly.

server/snapshot.go
```diff
@@ -234,6 +246,27 @@ func (r *raftFSM) Restore(snapshot io.ReadCloser) (retErr error) {
...
+				// s.raft.Snapshot() is actually s.raft.Raft.Snapshot()
+				// (our raft structure embeds Raft). But it is set after
+				// the call to NewRaft() - which invokes this function on
+				// startup. However, this runs under the server's lock, so
+				// simply grab the lock to ensure that Snapshot() will be
+				// able to execute.
+				// Also don't use startGoRoutine() here since it needs the
+				// server lock, which we already have.
+				s.mu.Lock()
+				s.raft.Snapshot().Error()
```
derekcollison: Can't do in place correct?

server/snapshot.go
```diff
@@ -325,43 +360,60 @@ func (r *raftFSM) restoreClientsFromSnapshot(serverSnap *spb.RaftSnapshot) error
...
 				break
 			}
 		}
+		if !ok {
+			err = fmt.Errorf("channel %q - unable to restore messages, aborting", sc.Channel)
+			s.log.Fatalf(err.Error())
+			// In tests, we use a "dummy" logger, so process will not exit
+			// (and we would not want that anyway), so make sure we return
+			// an error.
+			return false, err
```
derekcollison: Can we even get here with Fatalf?

server/snapshot.go
kozlovic: No, what is deferred is the invocation of the function. So function will execute and evaluate whatever is inside. I have tested it works ok.

server/snapshot.go
kozlovic: meaning without go routine? if so no, since as I tried to explain, as part of the startup process we call NewRaft() (hashicorp raft constructor), which returns a Raft object that we assign to s.raft.Raft at the end of createRaftNode(). NewRaft() is the one invoking Restore() on startup (when there are local snapshots), so I can't call s.raft.Snapshot() until after the creation of the raft node has completed.

server/snapshot.go
kozlovic: Read comment above ;-). Yes, we can and do in all tests since our test logger does not exit. It would be bad if we did because all tests that trigger a fatalf would cause the test suite to stop (due to call to os.Exit()).

server/snapshot.go
derekcollison: ok

PR #842, opened 2019-05-21
adopt project rename

MartinForReal:
rename go-nats to nats.go
rename go-nats-streaming to stan.go
rename project to fix unit test in stan.go

### DIFF
--- a/README.md
+++ b/README.md
@@ -2042,7 +2042,7 @@ if you want to ensure that each operation is immediately committed to the databa
 Here is the list of NATS Streaming clients, supported by Synadia. We may add additional supported streaming clients in the future, and encourage community-contributed clients.
 
 - [C#](https://github.com/nats-io/csharp-nats-streaming)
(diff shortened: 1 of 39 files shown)


kozlovic:
@MartinForReal We should not change the imports since the server uses vendor directory that still has code pre repo name change. Once NATS 2.0 is out, I may do a refresh of our dependencies and move to go.mod likely.
As of now, what could be changed is the link to clients in the README.md. You have done it for go-nats-streaming, but we have others clients repo names that have changed to.
So we should either close this PR or you can revert the import changes and maybe update the README with other client links? Of course, you don't have to and I can do it myself if you wish.
Thanks!

kozlovic:
@MartinForReal Actually, there is no need to update the links in the README.md. I have removed the list and added a link to the download page in nats-io instead (https://github.com/nats-io/nats-streaming-server/pull/844)

PR #867, opened 2019-07-01
Update NATS server vendor

derekcollison:
Signed-off-by: Derek Collison <derek@nats.io>


vtolstov:
why vendoring still needed if project already uses go.mod ?

derekcollison:
We are in a transition period. We just added go.mod support but also are still building with go 1.11.x for our releases and we force them to use vendor at the moment. We update vendor now with `go mod vendor` however and in a future commit you will see the vendor directory removed. Most likely once we move to go 1.12.x or above for release builds.

vtolstov:
if i'm use go.mod does this mean that i can simple go get github.com/nats-io/nats-streaming-serve@master and use all up to date deps automatic?

derekcollison:
It will get the master version yes. The dependencies that NATS streaming actually use those will not be updated IIRC>

kozlovic:
@vtolstov The go.mod in nats streaming should allow you to build the server with all its dependencies, yes. The vendor directory is built out of that and we can build the server from the vendor dir.

vtolstov:
thanks!

PR #875, opened 2019-07-04
[ADDED] FileStore: read ahead capability

kozlovic:
A new option `file_read_buffer_size` is added and defaults to 2MB.
If not set to 0, when a message is read from disk, the store will
actually read the following messages, up to that buffer size, and
add them to the internal cache (which already existed and has a
default TTL of 1sec).
This will speed up lookups of following messages since they are
likely to be in the cache.

### DIFF
--- a/server/conf.go
+++ b/server/conf.go
@@ -498,6 +498,11 @@ func parseFileOptions(itf interface{}, opts *Options) error {
 				return err
 			}
 			opts.FileStoreOpts.ParallelRecovery = int(v.(int64))
+		case "file_read_buffer_size", "read_buffer_size":
+			if err := checkType(k, reflect.Int64, v); err != nil {
+				return err
+			}
+			opts.FileStoreOpts.ReadBufferSize = int(v.(int64))
 		}
 	}
 	return nil
@@ -592,6 +597,7 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 	fs.IntVar(&sopts.FileStoreOpts.CompactInterval, "file_compact_interval", stores.DefaultFileStoreOptions.CompactInterval, "stan.FileStoreOpts.CompactInterval")
 	fs.String("file_compact_min_size", fmt.Sprintf("%v", stores.DefaultFileStoreOptions.CompactMinFileSize), "stan.FileStoreOpts.CompactMinFileSize")
 	fs.String("file_buffer_size", fmt.Sprintf("%v", stores.DefaultFileStoreOptions.BufferSize), "stan.FileStoreOpts.BufferSize")
+	fs.String("file_read_buffer_size", fmt.Sprintf("%v", stores.DefaultFileStoreOptions.ReadBufferSize), "")
 	fs.BoolVar(&sopts.FileStoreOpts.DoCRC, "file_crc", stores.DefaultFileStoreOptions.DoCRC, "stan.FileStoreOpts.DoCRC")
 	fs.Int64Var(&sopts.FileStoreOpts.CRCPolynomial, "file_crc_poly", stores.DefaultFileStoreOptions.CRCPolynomial, "stan.FileStoreOpts.CRCPolynomial")
 	fs.BoolVar(&sopts.FileStoreOpts.DoSync, "file_sync", stores.DefaultFileStoreOptions.DoSync, "stan.FileStoreOpts.DoSync")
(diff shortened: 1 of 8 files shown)


stores/filestore.go
```diff
@@ -3208,6 +3241,148 @@ func (ms *FileMsgStore) lookup(seq uint64) (*pb.MsgProto, error) {
...
+		}
+	}
+
+	// Position the data file to the offset of the first message we need to read.
+	file := fslice.file.handle
+	if _, err := file.Seek(firstMsgOffset, io.SeekStart); err != nil {
+		return nil, err
+	}
+	// Use a buffer that can hold all the message(s)
+	ms.tmpMsgBuf = util.EnsureBufBigEnough(ms.tmpMsgBuf, totalMsgsSize)
```
derekcollison: Do we reuse the via sync.Pool? Should we? Them and the index structs maybe?

stores/filestore.go
```diff
@@ -3401,6 +3576,7 @@ func (c *msgsCache) evict(now int64) {
 		// Bulk remove
 		c.seqMaps = make(map[uint64]*cachedMsg)
 		c.head, c.tail, c.tryEvict = nil, nil, 0
+		runtime.GC()
```
derekcollison: What does this block?

derekcollison: This should be fairly good these days with golang, not sure this is needed.

server/conf.go
```diff
@@ -690,10 +696,14 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp,
 			sopts.MaxBytes, flagErr = getBytes(f)
 		case "file_compact_min_size":
 			sopts.FileStoreOpts.CompactMinFileSize, flagErr = getBytes(f)
-		case "file_buffer_size":
+		case "file_buffer_size", "file_read_buffer_size":
```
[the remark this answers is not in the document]
kozlovic: I can separate them, it was just to avoid repeating this code ;-)
```
var i64 int64
i64, flagErr = getBytes(f)
```
... which is silly. Will fix.

stores/filestore.go
```diff
@@ -183,6 +184,11 @@ type FileStoreOptions struct {
 	// removed (the file is truncated at the beginning of the first incomplete
 	// record). Dataloss may occur.
 	TruncateUnexpectedEOF bool
+
+	// ReadBufferSize, if non zero, will cause the store to preload messages
+	// (up to this total size) when looking up a message. This is speeding
```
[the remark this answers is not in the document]
kozlovic: Will fix.

stores/filestore.go
```diff
@@ -3208,6 +3241,148 @@ func (ms *FileMsgStore) lookup(seq uint64) (*pb.MsgProto, error) {
...
+	maxIdxs := int(fslice.lastSeq - seq + 1)
+	done := false
+	idxBufSize := cap(idxsBuf)
+	for iter := 0; !done; iter++ {
+		if (iter+1)*128 > maxIdxs {
+			idxBufSize = (maxIdxs - (iter * 128)) * msgIndexRecSize
+			// Will be done after reading this page of indexes
+			done = true
+		}
+		b, err := io.ReadFull(fslice.idxFile.handle, idxsBuf[:idxBufSize])
```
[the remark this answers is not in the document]
kozlovic: An index on disk is 32 bytes right now. So in memory it would be 8 bytes for key, 20 for content (since we would not need the CRC - 4 bytes - that is on disk. I could make this another option to allow keeping channel message's index in memory..

Regardless, I wanted to revisit the reading of indexes in this read ahead code (that was not my original approach). Maybe I should use the same buffer size than for messages and read up to that number (or up to the number of indexes in this file slice) instead of reading by pages.

stores/filestore.go
```diff
@@ -3208,6 +3241,148 @@ func (ms *FileMsgStore) lookup(seq uint64) (*pb.MsgProto, error) {
...
+	msgSizes := _msgSizes[:0]
+
+	// Keep track of total number of messages and size
+	totalMsgs := 0
+	totalMsgsSize := 0
+	firstMsgOffset := int64(0)
+
+	// Indexes are fixed size. Read a page (4K) worth of them.
+	// Loop if needed if total message size is below readBufSize.
+	idxsBuf := [128 * msgIndexRecSize]byte{}
```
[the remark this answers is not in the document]
kozlovic: May go away if I change the way I read the indexes

stores/filestore.go
kozlovic: So ms.tmpMsgBuf is a FileMsgStore member and was used so far for reading/writing single messages.
I will probably replace with dedicated buffer (I have one for write calls) or as you say use a sync.Pool, which may be better/simpler than the implementation I had for shrinking the write buffer...

stores/filestore.go
kozlovic: I put that at one point because during some test bench, the mem usage was going quite high and so I used `-profile` and looked at memory in use and it was high. After adding this call, all mem usage for Unmarshal and add to cache disappeared, so that convinced me that I did not have a mem leak. So I will remove but the GC is not aggressive because that inuse_space stays for several minutes at the end of a test...

stores/filestore.go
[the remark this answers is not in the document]
derekcollison: I think if we have an active sub, up to some limit, we should keep the whole index for the channel past the floor (all active subs lowest seqno) in memory and avoid file IO.

stores/filestore.go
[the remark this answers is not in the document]
derekcollison: Still would make it a const or a var if its size of a struct or something, etc..

stores/filestore.go
derekcollison: Sounds good.

stores/filestore.go
derekcollison: ok just concerned this does a block here inline which we should avoid if we can.

PR #886, opened 2019-07-22
Fixed implementation of FileStore auto-sync feature

kozlovic:
The original implementation was having a central timer routine
that would go through the list of channels (interface) and cast
them to FileMsgStore and FileSubStore to invoke the stores' autoSync()
functions.
However, with stores that wrap a FileStore instance (such as RAFT or
Crypto store), that would cause a panic since the casting would fail.

Alternative fix would have been for FileStore to maintain a separate
map of actual File[Msg/Sub]Store, but I chose instead to move the
call to autoSync() into the Msg/Sub stores themselves.

Relates to #882

### DIFF
--- a/stores/filestore.go
+++ b/stores/filestore.go
@@ -510,7 +510,6 @@ type FileStore struct {
 	cliCompactTS  time.Time
 	crcTable      *crc32.Table
 	lockFile      util.LockFile
-	autoSyncTimer *time.Timer
 }
 
 type subscription struct {
@@ -546,6 +545,9 @@ type FileSubStore struct {
 	activity    bool         // was there any write between two flush calls
 	writer      io.Writer    // this is either `bw` or `file` depending if buffer writer is used or not
 	shrinkTimer *time.Timer  // timer associated with callback shrinking buffer when possible
+	syncTimer   *time.Timer  // timer associated with performing auto flush and disk sync
+	needSync    bool         // this is required to reduce sync'ing in case DoSync==false, but AutoSync>0
+	synced      int64        // number of times the file is actually sync'ed
 	allDone     sync.WaitGroup
 }
 
@@ -636,6 +638,8 @@ type FileMsgStore struct {
 	bkgTasksWake chan bool // signal the background tasks go routine to get out of a sleep
 	allDone      sync.WaitGroup
 	readBufSize  int
+	needSync     bool  // this required to reduce sync'ing when DoSync==false, but AutoSync>0
+	synced       int64 // number of times the file is actually sync'ed
 }
 
 type bufferPool struct {
@@ -669,7 +673,6 @@ var (
 	bkgTasksSleepDuration = defaultBkgTasksSleepDuration
 	cacheTTL              = int64(defaultCacheTTL)
 	sliceCloseInterval    = defaultSliceCloseInterval
-	testAutoSync          = int64(0)
 )
 
 // FileStoreTestSetBackgroundTaskInterval is used by tests to reduce the interval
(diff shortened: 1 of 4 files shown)


stores/filestore.go
```diff
@@ -546,6 +545,9 @@ type FileSubStore struct {
 	activity    bool         // was there any write between two flush calls
 	writer      io.Writer    // this is either `bw` or `file` depending if buffer writer is used or not
 	shrinkTimer *time.Timer  // timer associated with callback shrinking buffer when possible
+	syncTimer   *time.Timer  // timer associated with performing auto flush and disk sync
+	needSync    bool         // this is required to reduce sync'ing in case DoSync==false, but AutoSync>0
+	synced      int64        // number of times the file is actually sync'ed
```
derekcollison: Why do we need this? Just curious.

stores/filestore.go
```diff
@@ -3893,6 +3871,12 @@ func (fs *FileStore) newFileSubStore(channel string, limits *SubStoreLimits, doR
 	} else {
 		fs.fm.unlockFile(ss.file)
 	}
+	if fs.opts.AutoSync > 0 {
+		ss.Lock()
+		ss.allDone.Add(1)
+		ss.syncTimer = time.AfterFunc(fs.opts.AutoSync, ss.autoSync)
```
derekcollison: What if already set?

stores/filestore.go
```diff
@@ -4379,10 +4368,15 @@ func (ss *FileSubStore) Flush() error {
 func (ss *FileSubStore) autoSync() {
 	ss.Lock()
 	if !ss.closed {
-		if err := ss.lockFile(); err == nil {
-			ss.flush(true)
-			ss.fm.unlockFile(ss.file)
+		if ss.needSync {
+			if err := ss.lockFile(); err == nil {
```
derekcollison: Combine?

```go
if ss.needSync && ss.lockFile() == nil```

File: stores/filestore_test.go [hunk 4]
```diff
@@ -2001,50 +2017,60 @@ func TestFSAutoSync(t *testing.T) {
...
+		storeMsg(t, c, cn, 1, []byte("hello"))
+		subID := storeSub(t, c, cn)
+		storeSubDelete(t, c, cn, subID)
+	}
 
-	// Now delete channel
-	if err := s.DeleteChannel("foo"); err != nil {
-		t.Fatalf("Error deleting channel: %v", err)
+	// Delete channels around the time the auto-sync would hit.
+	time.Sleep(140 * time.Millisecond)
```
derekcollison: Feels too random for testing, but maybe its ok.

stores/filestore.go
kozlovic: You mean `synced` count? This is mainly to check in tests that we don't sync when not needed.

stores/filestore.go
kozlovic: Can't. This is the creation of the substore. For a given substore, this happen only once.

stores/filestore.go
kozlovic: ok

stores/filestore_test.go
kozlovic: To be honest, this test about DeleteChannel is not really relevant now that I don't do it at the FileStore level. Original code would grab all channels under FileStore lock, then release the lock and then individually auto-sync on each msg/sub store for each channel. The "risk" was that the channel had been deleted between the gather and the actual auto-sync. This can no longer happen since the auto-sync is brought down to the MsgStore and SubStore.

stores/filestore.go
derekcollison: ok

stores/filestore.go
derekcollison: ok

stores/filestore_test.go
derekcollison: Maybe delete the test then?

kozlovic: that part of it yes.

PR #927, opened 2019-08-30
added credentials file option for external nats server

david-gurley:


server/server.go
```diff
@@ -1253,36 +1253,37 @@ func (ss *subStore) LookupByAckInbox(ackInbox string) *subState {
...
+	Debug                 bool          // Debug trace
+	HandleSignals         bool          // Should the server setup a signal handler (for Ctrl+C, etc...)
+	Secure                bool          // Create a TLS enabled connection w/o server verification
+	ClientCert            string        // Client Certificate for TLS
+	ClientKey             string        // Client Key for TLS
+	ClientCA              string        // Client CAs for TLS
+	IOBatchSize           int           // Maximum number of messages collected from clients before starting their processing.
+	IOSleepTime           int64         // Duration (in micro-seconds) the server waits for more message to fill up a batch.
+	NATSServerURL         string        // URL for external NATS Server to connect to. If empty, NATS Server is embedded.
+	NATSServerCredentials string     // Credentials file for connecting to external NATS Server.
```
kozlovic: Let's name this simply`NATSCredentials` and make sure you do a `gofmt` (notice the `// Credentials..` was not aligned, this cause the build to fail).

server/server.go
```diff
@@ -1302,16 +1303,17 @@ func (o *Options) Clone() *Options {
...
-	ClientHBTimeout:   DefaultClientHBTimeout,
-	ClientHBFailCount: DefaultMaxFailedHeartBeats,
+	ID:                    DefaultClusterID,
+	DiscoverPrefix:        DefaultDiscoverPrefix,
+	StoreType:             DefaultStoreType,
+	FileStoreOpts:         stores.DefaultFileStoreOptions,
+	IOBatchSize:           DefaultIOBatchSize,
+	IOSleepTime:           DefaultIOSleepTime,
+	NATSServerURL:         "",
+	NATSServerCredentials: "",
```
kozlovic: Not needed (since default string is empty). We could even remove `NATSServerURL` from there..

server/server.go
```diff
@@ -1433,6 +1435,10 @@ func (s *StanServer) createNatsClientConn(name string) (*nats.Conn, error) {
 	ncOpts.Password = s.natsOpts.Password
 	ncOpts.Token = s.natsOpts.Authorization
 
+	if s.opts.NATSServerCredentials != "" {
```
kozlovic: Let's check that this is not already set with possibly the option passed programmatically with s.opts.NATSClientOpts (for people embedding the server). So maybe move this before line `for _, o := range s.opts.NATSClientOpts {` so that these options take precedence over the config file?

kozlovic:
@david-gurley I would want to do a release soon, so I used your branch and made the suggested modifications. The PR is #941 and will have the history of your commit for proper credits. Closing this PR now.

PR #931, opened 2019-09-03
[FIXED] Ensure Subscription response sent prior to messages

### LINKED ISSUE #930: Messages for durables possibly sent before Subscription response
If a durable subscription is closed (not unsubscribed) and later restarted while new messages are being published, it is possible that messages are sent prior to the subscription response protocol.
This is not visible in most of the client libraries since the subscription create usually holds a lock that is required during processing of the incoming messages.
However, this is not the case in Node.js.

kozlovic:
Original issue was addressed in PR #247, but unfortunately the fix
did not take into account durable subscriptions. If a durable
subscription is restarted, the subscription object is in memory
and would have the flag "initialized" already set to true, which
would not prevent published messages at the time of the subscription
request processing to allow messages to be sent prior to the
subscription response.

Resolves #930

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -4584,18 +4584,16 @@ func (s *StanServer) addSubscription(ss *subStore, sub *subState) error {
 }
 
 // updateDurable adds back `sub` to the client and updates the store.
-// No lock is needed for `sub` since it has just been created.
-func (s *StanServer) updateDurable(ss *subStore, sub *subState) error {
-	// Reset the hasFailedHB boolean since it may have been set
-	// if the client previously crashed and server set this
-	// flag to its subs.
-	sub.hasFailedHB = false
+func (s *StanServer) updateDurable(ss *subStore, sub *subState, clientID string) error {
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -4584,18 +4584,16 @@ func (s *StanServer) addSubscription(ss *subStore, sub *subState) error {
 }
 
 // updateDurable adds back `sub` to the client and updates the store.
-// No lock is needed for `sub` since it has just been created.
-func (s *StanServer) updateDurable(ss *subStore, sub *subState) error {
-	// Reset the hasFailedHB boolean since it may have been set
-	// if the client previously crashed and server set this
-	// flag to its subs.
-	sub.hasFailedHB = false
```
kozlovic: This change is not related to the `initialized` reset issue, but I moved that because with the updated test, I got a DATA race report. The actual fix for issue 930 is in processSub() by setting `initialized=false` for a recovered durable subscription.

PR #937, opened 2019-09-05
[IMPROVED] Pacing of subscription start

kozlovic:
When a subscription has a high MaxInflight and the subscription
starts, it forces the server to load messages up to MaxInflight
and deliver them. This can be taxing on the server.
Adding pacing to the subscription start. This also benefits to
queue subscriptions of the same group starting at the same time,
helping messages to be better distributed.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -125,6 +125,12 @@ const (
 
 	// Log statement printed when server is considered ready
 	streamingReadyLog = "Streaming Server is ready"
+
+	// Number of messages to send to a sub/queue on start of the subscription.
+	// This is to avoid having the server load a large amount of messages
+	// when the subscription start. It also allows queue subs starting
+	// at the same time to better distribute load.
+	defaultPaceLimit = 100
 )
 
 // Constant to indicate that sendMsgToSub() should check number of acks pending
@@ -181,6 +187,7 @@ var (
 	lazyReplicationInterval    = defaultLazyReplicationInterval
 	testDeleteChannel          bool
 	testSubSentAndAckSlowApply bool
+	paceLimit                  = defaultPaceLimit
 )
 
 var (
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -5002,11 +5009,17 @@ func (s *StanServer) processSubscriptionsStart() {
 				s.performDurableRedelivery(c, sub)
 			}
 			// publish messages to this subscriber
-			if qs != nil {
-				s.sendAvailableMessagesToQueue(c, qs)
-			} else {
-				s.sendAvailableMessages(c, sub)
-			}
+			s.wg.Add(1)
+			go func() {
```
derekcollison: Since we are pacing why also put into its own Go routine?

server/server.go
```diff
@@ -5133,46 +5146,78 @@ func (s *StanServer) processAck(c *channel, sub *subState, sequence uint64, from
...
+	i := 0
+	for nextSeq := qs.lastSent + 1; qs.stalledSubCount < len(qs.subs) && i < limit; nextSeq++ {
 		nextMsg := s.getNextMsg(c, &nextSeq, &qs.lastSent)
 		if nextMsg == nil {
 			break
 		}
 		if _, sent := s.sendMsgToQueueGroup(qs, nextMsg, honorMaxInFlight); !sent {
 			break
 		}
+		i++
```
derekcollison: Can you put this on end of the for statement?

server/server.go
```diff
@@ -5133,46 +5146,78 @@ func (s *StanServer) processAck(c *channel, sub *subState, sequence uint64, from
...
+	i := 0
+	for nextSeq := sub.LastSent + 1; !sub.stalled && i < limit; nextSeq++ {
 		nextMsg := s.getNextMsg(c, &nextSeq, &sub.LastSent)
 		if nextMsg == nil {
 			break
 		}
 		if sent, sendMore := s.sendMsgToSub(sub, nextMsg, honorMaxInFlight); !sent || !sendMore {
 			break
 		}
+		i++
```
derekcollison: Same question here. Also can these two funcs, qs and non qs be combined?

kozlovic:
> Have we seen if this helps in overall throughput when we have a lot of messages and a sub starts? 

We have seen that higher MaxInflight helps a consumer perf number. The idea of this was to reduce the amount of time/resource the server is "locked" in processing a subscription start with high MaxInflight.
The processing of a subscription start is already handled by a separate go routine which avoids the "timeout" that we would get if several subscriptions, each with a high MaxInflight and/or slow store, were to start at the same time. So it does not really help there. Where it helped was to have a more fair distribution on startup with queue members and also allow a subscription close to be processed more promptly. For instance say that you start a subscription with MaxInflight of 100,000 (and there are that many or more messages available), but the subscription stops after receiving say 100 messages. The server will still have the sub's lock to read and send 100,000 messages, preventing the processing of the subscription close.

I am going to close the PR for now and revisit later.

PR #943, opened 2019-09-18
reset hitLimit when Store

andyxning:
This PR resets `hitLimit` when Store. This will reset `hitLimit` after the limit is reached and this will once the limit is reached again a warning log will be recorded.

### DIFF
--- a/stores/filestore.go
+++ b/stores/filestore.go
@@ -2969,6 +2969,9 @@ func (ms *FileMsgStore) enforceLimits(reportHitLimit, lockFile bool) error {
 			return nil
 		}
 		if reportHitLimit && !ms.hitLimit {
+			defer func() {
+				ms.hitLimit = false
+			}()
 			ms.hitLimit = true
 			ms.log.Warnf(droppingMsgsFmt, ms.subject, ms.totalCount, ms.limits.MaxMsgs,
 				util.FriendlyBytes(int64(ms.totalBytes)), util.FriendlyBytes(ms.limits.MaxBytes))
--- a/stores/memstore.go
+++ b/stores/memstore.go
@@ -119,6 +119,9 @@ func (ms *MemoryMsgStore) Store(m *pb.MsgProto) (uint64, error) {
 				(maxBytes > 0 && (ms.totalBytes > uint64(maxBytes)))) {
 			ms.removeFirstMsg()
 			if !ms.hitLimit {
+				defer func() {
+					ms.hitLimit = false
+				}()
 				ms.hitLimit = true
 				ms.log.Warnf(droppingMsgsFmt, ms.subject, ms.totalCount, ms.limits.MaxMsgs,
 					util.FriendlyBytes(int64(ms.totalBytes)), util.FriendlyBytes(ms.limits.MaxBytes))
--- a/stores/sqlstore.go
+++ b/stores/sqlstore.go
@@ -1448,6 +1448,9 @@ func (ms *SQLMsgStore) Store(m *pb.MsgProto) (uint64, error) {
 				ms.first++
 			}
 			if !ms.hitLimit {
+				defer func() {
+					ms.hitLimit = false
+				}()
 				ms.hitLimit = true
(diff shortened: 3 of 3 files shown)


kozlovic:
@andyxning The point of the store's `hitLimit` was to print the warning only once. As you can see, with your defer you are just reseting the hitLimit to false just after it is set to true, making that boolean useless.

The reason this boolean exists is that in NATS Streaming, messages are never removed by consumers, only due to limits. So when a message is stored and causes the server to remove old message(s) it is going to happen for possibly every new produced message and would fill up the log with that.

I have to reject this PR, hope you understand.

andyxning:
> The reason this boolean exists is that in NATS Streaming, messages are never removed by consumers, only due to limits. So when a message is stored and causes the server to remove old message(s) it is going to happen for possibly every new produced message and would fill up the log with that.

I understood. But, imho, removing old messages is a crucial operation and this needs to be recorded. This can help us to investigate the status of the cluster. Even this will happen when every message comes in, the log should recorded.

 > The point of the store's hitLimit was to print the warning only once. As you can see, with your defer you are just reseting the hitLimit to false just after it is set to true, making that boolean useless.

By resetting the `hitLimit` value to `false` will make the log check when every message comes in not every message nats-streaming-server removes.

kozlovic:
But, unless you have expiration and some messages are expired (which do not trigger this warning), any *new* message will cause a warning. What's the difference between getting the warning once and for every single new stored message? If you are not expecting any warning due to store limits, I am not sure what difference it makes to have 1 warning or a million, they will all be the same, say:
```
[76175] 2019/09/19 17:11:11.098661 [WRN] STREAM: Reached limits for store "foo" (msgs=1000/1000 bytes=146.48 KB/976.56 MB), dropping old messages to make room for new ones
[76175] 2019/09/19 17:11:11.098664 [WRN] STREAM: Reached limits for store "foo" (msgs=1000/1000 bytes=146.48 KB/976.56 MB), dropping old messages to make room for new ones
[76175] 2019/09/19 17:11:11.098668 [WRN] STREAM: Reached limits for store "foo" (msgs=1000/1000 bytes=146.48 KB/976.56 MB), dropping old messages to make room for new ones
[76175] 2019/09/19 17:11:11.098672 [WRN] STREAM: Reached limits for store "foo" (msgs=1000/1000 bytes=146.48 KB/976.56 MB), dropping old messages to make room for new ones
[76175] 2019/09/19 17:11:11.098676 [WRN] STREAM: Reached limits for store "foo" (msgs=1000/1000 bytes=146.48 KB/976.56 MB), dropping old messages to make room for new ones
[76175] 2019/09/19 17:11:11.098680 [WRN] STREAM: Reached limits for store "foo" (msgs=1000/1000 bytes=146.48 KB/976.56 MB), dropping old messages to make room for new ones
[76175] 2019/09/19 17:11:11.098684 [WRN] STREAM: Reached limits for store "foo" (msgs=1000/1000 bytes=146.48 KB/976.56 MB), dropping old messages to make room for new ones
```

andyxning:
> unless you have expiration and some messages are expired (which do not trigger this warning), any new message will cause a warning.

This is quite true. Can we add an option for logging warning record about this and by default it is disabled?

kozlovic:
You mean an option to log when messages are expired?

I think you are missing the point of NATS Streaming. By definition messages are not removed when they are consumed. The only way for messages to be removed is due to limits. Setting a message expiration ensures that you can maintain some limits and keep the channel content below this limit.
Without expiration, you would be guaranteed to hit the limit just by the fact of using the system. Regardless of how high the limit is, if there is no removal of messages due to expiration, you will hit the limit at one point, and I don't think it makes sense to log that messages are being removed any time a new message is stored.

Again, message expiration is the only way to keep the channel content within some set limits and should not be reported as a warning when being expired in my opinion, since you have set the expiration and expect the messages to expire, it should not come as a surprise.

I could add a way to always report when older messages are removed to make room for the new ones, but even then, I would like you to convince me of the value of that and give me a concrete example on how having logging for each versus one makes a difference. Thanks!

andyxning:
@kozlovic OK. I learned. Thanks for your detailed explanation and patience. No more questions for now.

Issue #945, opened 2019-09-29
question about sequence numbers


biskit:
I have the following setup:

standalone NATS streaming server `STREAM: Starting nats-streaming-server[test-cluster] version 0.12.0`

and `Starting nats-server version 1.4.1`

and I start it like this: `./nats-streaming-server -st file -dir . -mi 5m -ma 1h -m 8222`

When I monitor the queues, the sequence number seems to sometimes go to 0. I thought the numbers never looped as I start my app with the last sequence number I retain.

For eg:
```
    {
      "name": "q1",
      "msgs": 169,
      "bytes": 3026271,
      "first_seq": 1,
      "last_seq": 169,
      "subscriptions": [
        {
          "client_id": "B6ECF42E28E954C39B4893240B2E870D_8161",
          "inbox": "_INBOX.eNuIfNNPnvhSGmYEU7SS47",
          "ack_inbox": "_INBOX.UhFI7O49QYeWsJQk0qgke3",
          "durable_name": "B6ECF42E28E954C39B4893240B2E870D_8161_q1_durable",
          "is_durable": true,
          "is_offline": false,
          "max_inflight": 1024,
          "ack_wait": 30,
          "last_sent": 169,
          "pending_count": 0,
          "is_stalled": false
        }
      ]
    }
```

I assumed the `first_seq`, `last_seq` will always be increasing till i delete my store?
My restarts depend on continuous increasing sequence numbers.

Or, is this sequence number different from `Msg.sequence`?

aricart:
From the usage and the doc, the `-mi,  --max_inactivity <duration>    Max inactivity (no new message, no subscription) after which a channel can be garbage collected (0 for unlimited)` implies that the channel itself is garbage collected. That means that the channel actually goes away if there are no subscriptions or new messages arriving - in your case this is set to 5 minutes.

Also if you are manually keeping track of the last sequence you saw, likely in this case where the channel is destroyed, I would expect that `StartAtTimeDelta` or `StartAtTime` would probably be safer.

kozlovic:
> I assumed the first_seq, last_seq will always be increasing till i delete my store?

And they are. But that's the key point, as @aricart mentioned, you are using max inactivity which causes the server to delete the channel store after that amount of time without new message and active consumers. I suspect that you have stopped your consumer application at one point, long enough for the server to decide to delete the channel.
If you want to keep message sequence numbers, and since you are already using message expiration, you should remove the `-mi` setting.

[...]

PR #955, opened 2019-10-21
[FIXED] The pod of nats-streaming will change status to “Completed” after k8s node reboot #954

### LINKED ISSUE #954: The pod of nats-streaming will change status to “Completed”  after k8s node reboot
After the k8s node which the pod of nats-streaming in rebooted，the status of nats-streaming pod will change to “Completed” for ever。The nats-streaming cluster is not ready for ever 

image

/kind bug

wanjunlei:
The reason of this error is when the nats streaming pod received a SIGTERM, the container of nats streaming server will exit with code 0. When container exit with code 0, K8s consider the container is noraml exit, so it will change the pods status to “Completed” and will not restart the container. If the container exit with code none zore, K8s will change the pods status to “Error” and restart the container.
  This pr change the exit code from 0 to 2 when nats streaming server received a SIGTERM. This can avoid the error of issue #945.

Resolves #954

### DIFF
--- a/server/signal.go
+++ b/server/signal.go
@@ -39,9 +39,12 @@ func (s *StanServer) handleSignals() {
 			// registered, so we don't need a "default" in the
 			// switch statement.
 			switch sig {
-			case syscall.SIGINT, syscall.SIGTERM:
+			case syscall.SIGINT:
 				s.Shutdown()
 				os.Exit(0)
+			case syscall.SIGTERM:
+				s.Shutdown()
+				os.Exit(2)
 			case syscall.SIGUSR1:
 				// File log re-open for rotating file logs.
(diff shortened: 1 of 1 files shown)


kozlovic:
@wanjunlei Thank you for the contribution! I understand the issue, but I wonder if you should return a different error code than 2? For instance, if the process was not trapping SIGTERM, the exit code reported would be 143. Could you change to this value?
Also, would you mind doing a rebase from master because there have been some fixes that without are preventing the CI integration. Thanks!

wanjunlei:
@kozlovic In that case，it only need a none zore exit code. I will change it to 143 and rebase from master. Thank you！

PR #973, opened 2019-11-15
Ifelseswitch

tegk:
I refactored several if-else statement to more readable switch statements.
Furthermore I refactored a single which statement to a if statement.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1411,14 +1411,14 @@ func (s *StanServer) buildServerURLs() ([]string, error) {
 		// use 127.0.0.1 or ::1 for host address (important for
 		// Windows since connect with 0.0.0.0 or :: fails).
 		sport := strconv.Itoa(opts.Port)
-    switch opts.Host {
-    case "0.0.0.0":
-      hostport = net.JoinHostPort("127.0.0.1", sport)
-    case "::", "[::]":
-      hostport = net.JoinHostPort("::1", sport)
-    default:
-      hostport = net.JoinHostPort(opts.Host, sport)
-    }
+		switch opts.Host {
+		case "0.0.0.0":
+			hostport = net.JoinHostPort("127.0.0.1", sport)
+		case "::", "[::]":
+			hostport = net.JoinHostPort("::1", sport)
+		default:
+			hostport = net.JoinHostPort(opts.Host, sport)
+		}
 	}
 	return []string{fmt.Sprintf("nats://%s", hostport)}, nil
 }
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -1411,11 +1411,12 @@ func (s *StanServer) buildServerURLs() ([]string, error) {
 		// use 127.0.0.1 or ::1 for host address (important for
 		// Windows since connect with 0.0.0.0 or :: fails).
 		sport := strconv.Itoa(opts.Port)
-		if opts.Host == "0.0.0.0" {
+		switch opts.Host {
+		case "0.0.0.0":
 			hostport = net.JoinHostPort("127.0.0.1", sport)
-		} else if opts.Host == "::" || opts.Host == "[::]" {
+		case "::":
```
ripienaar: `case "::", "[::]":` I imagine, else we are not functionally equiv here

tegk: true, will add

ripienaar:
good stuff, certainly looks more readable to me. LGTM but we need to get another maintainer to chime in too I suppose

server/server.go
```diff
@@ -1411,11 +1411,14 @@ func (s *StanServer) buildServerURLs() ([]string, error) {
...
 			hostport = net.JoinHostPort("::1", sport)
-		} else {
+		case "[::]":
```
kozlovic: I think what @ripienaar was saying is that you could have the `"::"` and `[::]` in the same case statement:
```
switch opts.Host {
    case "0.0.0.0":
	hostport = net.JoinHostPort("127.0.0.1", sport)
    case "::", "[::]":
	hostport = net.JoinHostPort("::1", sport)
    default:
	hostport = net.JoinHostPort(opts.Host, sport)
}
```

tegk: True, I changed it.

server/server.go
```diff
@@ -1411,13 +1411,14 @@ func (s *StanServer) buildServerURLs() ([]string, error) {
...
 		// Windows since connect with 0.0.0.0 or :: fails).
 		sport := strconv.Itoa(opts.Port)
-		if opts.Host == "0.0.0.0" {
-			hostport = net.JoinHostPort("127.0.0.1", sport)
-		} else if opts.Host == "::" || opts.Host == "[::]" {
-			hostport = net.JoinHostPort("::1", sport)
-		} else {
-			hostport = net.JoinHostPort(opts.Host, sport)
-		}
+    switch opts.Host {
```
kozlovic: You need to run "go fmt" because this is not properly formatted (see that our CI caught that).

PR #976, opened 2019-11-21
[ADDED] In/Out Msgs/Bytes in /serverz monitoring endpoint

### LINKED ISSUE #974: [API] field total_msgs (streaming/serverz) is directly affected by the deletion.
Hello guys

We use [nats-exporter](https://github.com/nats-io/prometheus-nats-exporter) to get nats metrics and set up charts, alerts, and everything.

We note that after the timeout to begin deleting messages, the`streaming/serverz` -> `total_msgs` field follows this deletion and matched its value relative to what is in storage. This is not good for taking real-time metrics of the amount of messages we receive in the streaming input.

The metrics are confusing as it takes into account what is being deleted as well:
`round(sum(rate(nss_chan_msgs_total{server_id="$server"}[5m])) by (instance))`
Screenshot_20191118_160830
(I'm not send this amount of messages)

Is there a possibility that we have a field for streaming like nats-server already has? `/ varz` ->` in_msgs`

Thank you!

kozlovic:
Unlike total_msgs/bytes that represent the total number of messages
stored, in_msgs/bytes and out_msgs/bytes are the messages received
by the server and sent to clients.
The `in_msgs` counter is updated when processing inbound client
messages.
The `out_msgs` counter is updated when sending a message to a subscription,
either as a new or redelivered message.

Note that the size for in_bytes/out_bytes is different because these
values represent the bytes received (a PubMsg protobuf) and sent
(a MsgProto protobuf) that have different overhead.

Resolves #974

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -21,6 +21,7 @@ import (
 	"runtime"
 	"sort"
 	"strconv"
+	"sync/atomic"
 	"time"
 
 	natsd "github.com/nats-io/nats-server/v2/server"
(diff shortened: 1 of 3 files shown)


kozlovic:
@ColinSullivan1 As requested from user in issue #974, I am adding in_msgs/in_bytes and out_msgs/out_bytes in `streaming/serverz` endpoint. This is similar to NATS core in/out msgs/bytes. Since the user is using the nats-exporter, I think that once this is merged, changes would have to be made in the nats-exporter to take advantage of those 4 new fields, right?

PR #978, opened 2019-11-25
Fix typo in usage for --encryption_key

derekbekoe:


nats-streaming-server.go
```diff
@@ -51,7 +51,7 @@ Streaming Server Options:
     -sl,  --signal <signal>[=<pid>]      Send signal to nats-streaming-server process (stop, quit, reopen)
           --encrypt <bool>               Specify if server should use encryption at rest
           --encryption_cipher <string>   Cipher to use for encryption. Currently support AES and CHAHA (ChaChaPoly). Defaults to AES
-          --encryption_key <sting>       Encryption Key. It is recommended to specify it through the NATS_STREAMING_ENCRYPTION_KEY environment variable instead
+          --encryption_key <string>       Encryption Key. It is recommended to specify it through the NATS_STREAMING_ENCRYPTION_KEY environment variable instead
```
kozlovic: Good catch. But since you added a missing letter, I would ask that you remove one of the "space" character before the "description" so that it is aligned with others. Thanks!

derekbekoe: fixed. thanks.

PR #1024, opened 2020-02-27
[ADDED] Monitoring endpoint '/isFTActive' to evaluate whether the server is in FTActive mode

mattharr-is:
My use case for this feature is related to a kubernetes based deployment. We run nats in FT mode, which means that the monitor endpoints return useful information, or no information, depending on which node you make requests against. This feature will enable me to place some kind of reverse proxy in front of the nats nodes to proxy http requests on the monitor port to only the active node. 

With this PR in its current state, the only proxy i've found that i can configure to use appropriately is envoy. it seems most of the common reverse proxy solutions in use on k8s have varying levels of opinionatedness about how to configure health checks, and what constitutes "healthy". For instance, traefik considers any health check that returns a 2XX or 3XX to be healthy, and there's no way to alter that configuration. NGINX Plus supports active health checks that allow you to specify a path to a health check endpoint. The free NGINX offering only supports passive health checks, which evaluate traffic as a whole and removes upstream nodes from load balancing if a node begins throwing 5XX status codes. Envoy seems to be only proxy that allows you to fully configure a health check that fits inline with the expected status codes added by this feature.

### DIFF
--- a/server/monitor_test.go
+++ b/server/monitor_test.go
@@ -377,43 +377,45 @@ func TestMonitorServerz(t *testing.T) {
 	monitorExpectStatus(t, ServerPath, http.StatusInternalServerError)
 }
 
-func TestMonitorIsFTActiveStandaloneServer(t *testing.T) {
-	expectedStatusCode := http.StatusNoContent
-
-	resetPreviousHTTPConnections()
-	s := runMonitorServer(t, GetDefaultOptions())
-	defer s.Shutdown()
-
-	resp, _ := getBodyEx(t, http.DefaultClient, "http", IsFTActivePath, expectedStatusCode, "")
-	defer resp.Body.Close()
-
-	if expectedStatusCode != resp.StatusCode {
-		t.Fatalf("Expected status code %v, got %v", expectedStatusCode, resp.StatusCode)
-	}
-}
-
 func TestMonitorIsFTActiveFTServer(t *testing.T) {
-	expectedStatusCode := http.StatusOK
+	for _, test := range []struct {
+		name           string
+		checkActive    bool
+		expectedStatus int
+	}{
+		{"active", true, http.StatusOK},
+		{"standby", false, http.StatusNoContent},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			resetPreviousHTTPConnections()
 
-	resetPreviousHTTPConnections()
+			cleanupFTDatastore(t)
+			defer cleanupFTDatastore(t)
 
-	ncOpts := natsdTest.DefaultTestOptions
-	ns := natsdTest.RunServer(&ncOpts)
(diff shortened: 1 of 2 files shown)


server/monitor_test.go
```diff
@@ -377,6 +377,46 @@ func TestMonitorServerz(t *testing.T) {
...
+	expectedStatusCode := http.StatusNoContent
+
+	resetPreviousHTTPConnections()
+	s := runMonitorServer(t, GetDefaultOptions())
+	defer s.Shutdown()
+
+	resp, _ := getBodyEx(t, http.DefaultClient, "http", IsFTActivePath, expectedStatusCode, "")
+	defer resp.Body.Close()
+
+	if expectedStatusCode != resp.StatusCode {
```
kozlovic: You don't need this check since getBodyEx() will already ensure that the response status matched the one you pass as a parameter.

server/monitor_test.go
```diff
@@ -377,6 +377,46 @@ func TestMonitorServerz(t *testing.T) {
...
+		t.Fatalf("Expected status code %v, got %v", expectedStatusCode, resp.StatusCode)
+	}
+}
+
+func TestMonitorIsFTActiveFTServer(t *testing.T) {
```
kozlovic: If you don't mind, we could replace this test with the one below because then we check both the active and standby server.
```
func TestMonitorIsFTActiveFTServer(t *testing.T) {
	for _, test := range []struct {
		name           string
		checkActive    bool
		expectedStatus int
	}{
		{"active", true, http.StatusOK},
		{"standby", false, http.StatusNoContent},
	} {
		t.Run(test.name, func(t *testing.T) {
			resetPreviousHTTPConnections()

			cleanupFTDatastore(t)
			defer cleanupFTDatastore(t)

			nopts := defaultMonitorOptions
			var aNOpts *natsd.Options
			var sNOpts *natsd.Options

			if test.checkActive {
				aNOpts = &nopts
			} else {
				sNOpts = &nopts
			}

			opts := getTestFTDefaultOptions()
			active := runServerWithOpts(t, opts, aNOpts)
			defer active.Shutdown()

			getFTActiveServer(t, active)

			opts = getTestFTDefaultOptions()
			opts.NATSServerURL = "nats://127.0.0.1:4222"
			standby := runServerWithOpts(t, opts, sNOpts)
			defer standby.Shutdown()

			resp, _ := getBodyEx(t, http.DefaultClient, "http", IsFTActivePath, test.expectedStatus, "")
			defer resp.Body.Close()
		})
	}
}
```

mattharr-is: hey @kozlovic I went ahead and replaced both tests with the one you described as it's much more concise and validates both codepaths. thanks for the contribution!

PR #1025, opened 2020-03-09
[FIXED] Windows Service: do not force usage of syslog

kozlovic:
When starting as a Windows Service, we were creating a syslog to
possibly log the failure to start. But users running without
admin privileges would be unable to use the service since the
server would fail to create the event log entry.
So create the syslog only if explicit set (-syslog <bool> option)
or if no logfile has been specified.


kozlovic:
This was reported through Google mailing list and was confirmed to be addressing the issue: https://groups.google.com/d/msg/natsio/d0Jk3N7iqZ8/Hcds3p5LAQAJ

PR #1070, opened 2020-07-15
    SQLStore: Channel expiration: Handle sql no rows error.

hackacharya:
Once we get into this scenario, looks like we will be on it for ever -
    Handle sql no rows error gracefully, do not reset the timer and avoid  
    the sqlnorows-timerreset  loop. We already have checks for 
    timestamp = 0, count=0 etc here in this function. 
    
    Example logs from a system that was overloading the log stream
    Looks like this went on for a while .. 

    2020/07/14 07:48:05.271047 [ERR] STREAM: Unable to perform expiration for channel "fd56375b": sql: error executing "SELECT timestamp FROM Messages WHERE id=$1 AND seq>=$2 ORDER BY seq LIMIT 1": sql: no rows in result set
    ...
    2020/07/14 10:16:42.839703 [ERR] STREAM: Unable to perform expiration for channel "fd56375b": sql: error executing "SELECT timestamp FROM Messages WHERE id=$1 AND seq>=$2 ORDER BY seq LIMIT 1": sql: no rows in result set

### DIFF
--- a/stores/sqlstore.go
+++ b/stores/sqlstore.go
@@ -1572,7 +1572,10 @@ func (ms *SQLMsgStore) expireMsgs() {
 	for {
 		expiredTimestamp := time.Now().UnixNano() - int64(ms.limits.MaxAge)
 		r := ms.sqlStore.preparedStmts[sqlGetExpiredMessages].QueryRow(ms.channelID, expiredTimestamp)
-		if err := r.Scan(&count, &maxSeq, &totalSize); err != nil {
+
+		// It has been observed that, there are cases where sqlErrNoRows
+		// is returned in this case we shouldn't reset the timer
+		if err := r.Scan(&count, &maxSeq, &totalSize); (err != nil) && (err != sql.ErrNoRows) {
 			processErr(sqlGetExpiredMessages, err)
 			return
 		}
--- a/stores/sqlstore.go
+++ b/stores/sqlstore.go
@@ -1573,9 +1573,7 @@ func (ms *SQLMsgStore) expireMsgs() {
 		expiredTimestamp := time.Now().UnixNano() - int64(ms.limits.MaxAge)
 		r := ms.sqlStore.preparedStmts[sqlGetExpiredMessages].QueryRow(ms.channelID, expiredTimestamp)
 
-		// It has been observed that, there are cases where sqlErrNoRows
-		// is returned in this case we shouldn't reset the timer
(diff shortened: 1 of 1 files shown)


kozlovic:
Thank you for the report and PR, however, I want to find out why we are getting there in the first place. Normally a timer is created only when a message is stored, and since we always keep a message in the channel, there should always be a message waiting to expire. Meaning that at worst, if the timer fires too soon (say 1ms before the actual first message timestamp) and the query does not find a message to expire, it should get it at the next try. So the error would not be printed continuously.

One case where I can see the query return an error is if I manually delete messages from the store, leaving the channel empty. But then, we get the same issue just few lines down when trying to get the first timestamp.

If you could share any specific configuration that you have for this channel and if you know what could have led to the error in the first place it would be greatly appreciated. We may end up with the solution you propose (or even possibly do this check for ErrNoRows in processErr itself), but I would feel better if I understand the root of the issue.

kozlovic:
It could be a difference in how Scan() is implemented and that in some libs it returns just count, maxSeq, totalSize as 0 but no error and in some it returns the error.
Would you mind sharing which NATS Streaming server version you are using and if you are building yourself or use a release. If you built yourself, are you using the vendor directory or have you updated the lib?
Which driver are you using, mysql or postgres?
Thanks!

kozlovic:
Actually, if you look at the error message:
```
2020/07/14 07:48:05.271047 [ERR] STREAM: Unable to perform expiration for channel "fd56375b": sql: error executing "SELECT timestamp FROM Messages WHERE id=$1 AND seq>=$2 ORDER BY seq LIMIT 1": sql: no rows in result set
```
it does not come from the Scan that you have change the error check, it comes from selecting the timestamp, which happens line 1592.

hackacharya:
Thanks for the response. 

You are correct, this fix is improper and/or doesn't address the real issue.  This does happens much further down. Overlooked that, this has not been reproduced so the fix was based on code review... and for  your other question. This happened in a dev environment - it is not entirely unlikely that DB  tables were touched, full shutdown restart may also have happened.  This was with 16.x  And yes it is built and used.   

Wondering if it is still a good idea to have the sql.ErrNoRows check the place where you are suggesting - since the timer will get reset. when a new message shows up the timer will be created right. May be able to run a specific test ... will update the PR

hackacharya:
May be the line numbers are off  sqlGetFirstMsgTimestamp is where this error is from addressed that.

PR #1075, opened 2020-07-28
[ADDED] ClientURL() API and expose an OS-assigned server port via the NATS server options 

adrian-thurston:
If a random port is asked for in a call to `RunServerWithOpts`, return this port
in the `*server.Options argument`. Replaces the Port field with the OS assigned
port.

This PR also adds `ClientURL()` to `StanServer` as an alternate method of
fetching connection information.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1816,6 +1816,8 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 
 // ClientURL returns the basic URL string representation suitable for a client to use to connect
 func (s *StanServer) ClientURL() string {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	if s.providedServerURL != "" {
 		return s.providedServerURL
 	} else if s.natsServer != nil {
--- a/server/server.go
+++ b/server/server.go
@@ -1763,6 +1763,9 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 		if err := s.startNATSServer(); err != nil {
 			return nil, err
 		}
+		if natsOpts != nil && natsOpts.Port == server.RANDOM_PORT {
+			natsOpts.Port = nOpts.Port
+		}
 	}
 	// Check for monitoring
 	if nOpts.HTTPPort != 0 || nOpts.HTTPSPort != 0 {
--- a/server/server.go
+++ b/server/server.go
@@ -1808,6 +1808,10 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 	return &s, nil
 }
 
+func (s *StanServer) ClientURL() string {
+	return s.natsServer.ClientURL()
+}
+
 // Logging in STAN
 //
 // The STAN logger is an instance of a NATS logger, (basically duplicated
--- a/server/server.go
+++ b/server/server.go
@@ -756,6 +756,9 @@ type StanServer struct {
 	// cause races when running more than 1 server in a program or test.
 	pingResponseOKBytes            []byte
 	pingResponseInvalidClientBytes []byte
+
+	// If using an external server, capture the URL that was given for return in ClientURL().
+	providedServerURL string
 }
 
 type subsSentAndAckReplication struct {
@@ -1442,6 +1445,9 @@ func (s *StanServer) buildServerURLs() ([]string, error) {
 		}
 		// Use net.Join to support IPV6 addresses.
 		hostport = net.JoinHostPort(host, port)
+
+		// Capture for ClientURL()
+		s.providedServerURL = natsURL
 	} else {
 		// We embed the server, so it is local. If host is "any",
 		// use 127.0.0.1 or ::1 for host address (important for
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -1490,6 +1490,9 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 		if err := s.startNATSServer(); err != nil {
 			return nil, err
 		}
+		if natsOpts.Port == server.RANDOM_PORT {
+			natsOpts.Port = nOpts.Port
```
ripienaar, line 1494: I wouldn't expect the server to change the supplied options even if it's a pointer, wouldn't an accessor function be a bit more user friendly? Or maybe something similar to the nats `ClientURL()`

adrian-thurston: Thanks for your review @ripienaar, I did it this way because the underlying library also does it this way. The URL method would also work for us. Whichever way you think is best suits us, so long as we can have the OS assigned port.

ripienaar: Ah, ok, well @kozlovic would be main stakeholder here, I just saw it go past and thought that was a bit odd

kozlovic: It is true that this is how NATS Server is doing and many NATS Server tests are using the updated options. However, maybe exposing a new API ClientURL() that would either return the underlying natsServer.ClientURL() or "" if connecting remotely, would be better. @adrian-thurston Would you feel like doing this or I can do it easily (and include a test). Just let me know.

adrian-thurston: Oh I'm sorry I was wrong ... we actually do use the port plainly, so having it come back in the options as an int means we don't need to parse it out of the client url.

kozlovic: @adrian-thurston To be sure i understand what you are saying: you would prefer the update of options as opposed to have a new ClientURL() api so that you have direct access to the port for some other use, right?

kozlovic: Need to replace this "if" statement with:
```
if natsOpts != nil && natsOpts.Port == server.RANDOM_PORT {
```
because as you can see from the CI runs, this otherwise causes a panic since natsOpts can be nil.
The streaming server accepts "nats options" that can be nil, in which case it uses default NATS options, which in this case is obviously not accessible to caller since caller passed nil. Even when provided, the server makes a clone so that it is owned and can't be changed underneath by the user.

I would also like you to add this test in `server_run_test.go`:
```
func TestServerRandomPort(t *testing.T) {
	natsOpts := natsdTest.DefaultTestOptions
	natsOpts.Port = natsd.RANDOM_PORT
	opts := GetDefaultOptions()
	s := runServerWithOpts(t, opts, &natsOpts)
	defer s.Shutdown()

	if natsOpts.Port == natsd.RANDOM_PORT {
		t.Fatal("port was not updated")
	}
	s.Shutdown()

	// Try with no nats options provided to make sure we don't
	// access natsOpts without checking that it is not nil
	s = runServerWithOpts(t, opts, nil)
	s.Shutdown()
}
```
Thanks!

adrian-thurston: That's right, I mean it doesn't need to come back in the options per se, but just to have it available as an int would be great.

For sure, I'll fix for the natsOpts == nil case, add the test and resubmit.

Thank you for your feedback!

ripienaar:
We could return a *net.URL which serves most needs including the easy access to the port.

kozlovic:
> We could return a *net.URL which serves most needs including the easy access to the port.

Actually `*url.URL.Port()` returns a string, so would still have to convert to `int` if `int` is what is needed by the user.
That being said, we could certainly do both (update `natsOpts.Port` and provide `ClientURL()` that returns an `*url.URL`).

server/server.go
```diff
@@ -1805,6 +1808,10 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 	return &s, nil
 }
 
+func (s *StanServer) ClientURL() string {
```
kozlovic, line 1811: Since this is exported, should have a comment:
```
// ClientURL returns the basic URL string representation suitable for a client to use to connect
```
But see the comment about when the NATS server is not embedded.

kozlovic (changes requested):
Sorry, still some changes to make, and an open question: what should ClientURL() return if the Streaming server is not embedding the NATS Server?

server/server_test.go
```diff
@@ -1370,3 +1373,46 @@ func TestAckProcessedBeforeClose(t *testing.T) {
...
+	s.Shutdown()
+}
+
+func TestServerRandomClientURL(t *testing.T) {
+	natsOpts := natsdTest.DefaultTestOptions
+	natsOpts.Port = natsd.RANDOM_PORT
+	opts := GetDefaultOptions()
+	s := runServerWithOpts(t, opts, &natsOpts)
+	clientURL := s.ClientURL()
+	defer s.Shutdown()
```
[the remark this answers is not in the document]
adrian-thurston: @kozlovic No worries, I'm not too familiar with this project and it is important to get it right. I just pushed a new commit in which any provided URL is stashed and used in the ClientURL() call. I think this is equivalent to Non-embedded case. If there is a better way please let me know.

kozlovic (changes requested):
Almost there :-)

server/server.go
```diff
@@ -1805,6 +1814,17 @@ func RunServerWithOpts(stanOpts *Options, natsOpts *server.Options) (newServer *
 	return &s, nil
 }
 
+// ClientURL returns the basic URL string representation suitable for a client to use to connect
+func (s *StanServer) ClientURL() string {
+	if s.providedServerURL != "" {
```
kozlovic, line 1819: Add this before this line:
```
s.mu.Lock()
defer s.mu.Unlock()
```

kozlovic (changes requested):
Sorry, the comment was not added when I submitted the review. Just add locking and we should be ok.

kozlovic (approved):
LGTM. Thank you for your contribution!

PR #1087, opened 2020-09-14
[FIXED] FileStore: index file may be recreated with wrong offsets

### LINKED ISSUE #1086: FileStore: possibly create index with wrong offsets after data/index mismatch
When the server recovers an index file, it makes sure that the last index information matches the message in the dat file. When this is not the case, it will regenerate the index file based on the content of the .dat file. However, it is possible that the server then creates this index with wrong offsets.

/cc @johnsusi

kozlovic:
This could happen if recovery of an index does not have matching
information with the last message in corresponding ".dat" file.
In that case, the server wipes the index file and recreate it
based on the data file. However, the offset was not reset, which
means that the index file would contain wrong offsets.

Resolves #1086


derekcollison (approved):
LGTM

PR #1088, opened 2020-09-15
[FIXED] Clustering: possible panic on restart with "log not found"

kozlovic:
Was found by chance when causing backend boltdb store to throw
error when storing the raft log. The fact that a front-end log
cache was used and improperly caching logs that had failed to
be stored caused an issue for raft library. I have filed an
issue (and PR to fix) to the hashicorp/raft repo, but actually
decided to not use their LogCache since it can be easily added
to our raftLog implementation, reducing locking.


derekcollison (approved):
LGTM

PR #1090, opened 2020-09-17
[ADDED] Clustering: ability to add/remove nodes at runtime

kozlovic:
If option `allow_add_remove_node` in `streaming{cluster{}}` block
is enabled (or command line `--cluster_allow_add_remove_node`)
then the leader will listen to `_STAN.raft.<cluster id>.node.remove`
(and `...node.add`) to which an user can send a simple NATS request
with the value being the name of the node to add/remove.

If successful, the leader will return `+OK`, if not `-ERR` with
the error string explaining why it failed.

Note that removing a non-existent node will still return `+OK`.

If the leader receives a request to remove itself, it will work
but the process will then exit. It can be added back by sending
the request to the `..node.add` subject.

Note that this updates the RAFT configuration in the RAFT log but
does NOT change the configuration file. The admin will be responsible
to change the configuration on file so that when nodes are restarted
they reflect the new configuration (list of peers).

### DIFF
--- a/nats-streaming-server.go
+++ b/nats-streaming-server.go
@@ -52,18 +52,19 @@ Streaming Server Options:
           --encrypt <bool>               Specify if server should use encryption at rest
           --encryption_cipher <string>   Cipher to use for encryption. Currently support AES and CHAHA (ChaChaPoly). Defaults to AES
           --encryption_key <string>      Encryption Key. It is recommended to specify it through the NATS_STREAMING_ENCRYPTION_KEY environment variable instead
-    
+
 Streaming Server Clustering Options:
-    --clustered <bool>                   Run the server in a clustered configuration (default: false)
-    --cluster_node_id <string>           ID of the node within the cluster if there is no stored ID (default: random UUID)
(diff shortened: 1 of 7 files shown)


rwenz3l (approved):
Awesome Stuff! 💯 

I reviewed the code so far and that looks all good to me. I will try to setup a docker-compose stack and a client later to do some practical testing and see how that works out.

UPDATE: I tested the new flag with a local cluster, not with docker, but just multiple instances, because that made it a lot easier to work with. I've put the documentation of my testing here:
→ https://gist.github.com/rwenz3l/1671dd77cbc3790af29b3b42266dc242

The results from my testing:
- The PubSub message is received correctly and the node is shutdown as well as removed from the (raft) cluster
- Other peers keep trying to connect to the cluster, until the configuration has been changed and reloaded.
- Re-starting the node will not add them back to the cluster, but keep them in the candidate state indefinitely, unless we add it back in. Once added, they turn from Candidate to Follower.

So this looks really promising! Thanks a lot.

derekcollison (approved):
LGTM

PR #1091, opened 2020-09-17
[IMPROVED] Clustering: Report possible misconfiguration of peers list

### LINKED ISSUE #990: Nats cluster - nodes are talking, but election is not happening
I have a 3 node NATS cluster.  fairly basic configuration.  The nodes seem to communicate, as the routing detail (for all 3 nodes) is visible when connecting to an individual node's web server.

However, I have enabled cluster_raft_logging at startup, and it clearly shows that the raft election is not happening.

what am I missing?  I have compared these configurations to some other dev/test clusters, and I can see no obvious differences.   appropriate firewall ports are opened on all 3 nodes,

[nss-node-a.txt](https://github.com/nats-io/nats-streaming-server/files/3952811/nss-node-a.txt)
[nss-node-b.txt](https://github.com/nats-io/nats-streaming-server/files/3952813/nss-node-b.txt)
[nss-node-c.txt](https://github.com/nats-io/nats-streaming-server/files/3952815/nss-node-c.txt)

kozlovic:
The clustering peers list is supposed to contain the list of peer
node ids. However, it happens that users make the mistake of listing
the node peer ids in the same string as such:
```
streaming {
  cluster {
    node_id: "a"
    peers: ["a, b, c"]
  }
}
```
The correct definition of `peers` should be:
```
   peers: ["a", "b", "c"]
```

The server will now list the cluster known peers on startup and
will print a warning if a comma is detected in a peer name:
```
[INF] STREAM: peer 1: "a,b,c"
[WRN] STREAM: Peer name contains ',' make sure you provided an array of peer names, not a string with commas
```

Resolves #990

### DIFF
--- a/server/clustering_test.go
+++ b/server/clustering_test.go
@@ -3801,6 +3801,39 @@ func TestClusteringNodeIDInPeersArray(t *testing.T) {
 	s1.Shutdown()
 }
 
+func TestClusteringNodeWrongPeerList(t *testing.T) {
+	cleanupDatastore(t)
+	defer cleanupDatastore(t)
+	cleanupRaftLog(t)
+	defer cleanupRaftLog(t)
+
+	ns := natsdTest.RunDefaultServer()
+	defer ns.Shutdown()
+
+	l := &captureWarnLogger{}
+
+	s1Opts := getTestDefaultOptsForClustering("a", false)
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #1100, opened 2020-10-08
[ADDED] Authentication fields in streaming configuration block

kozlovic:
Abiltity to provide Username, Password or Token in the streaming
configuration block.

If the configuration is:
```
authorization {
  users [
    { user: foo, password: bar}
  ]
}
streaming {
  ..
}
```
the server would fail to start unless user passes `-user foo -pass bar`
from the command line. But there was no way to pass those from the
configuration file. You can now configure this way:

```
streaming {
  ..
  username: "foo"
  password: "bar"
}
```
in order for the server to be able to start without the additional
`-user` and `-pass` command line parameters.

Also adding the `token` configuration option.

### DIFF
--- a/server/conf.go
+++ b/server/conf.go
@@ -179,6 +179,21 @@ func ProcessConfigFile(configFile string, opts *Options) error {
 			}
 			opts.Encrypt = true
 			opts.EncryptionKey = []byte(v.(string))
+		case "username", "user":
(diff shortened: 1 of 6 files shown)


derekcollison (approved):
LGTM

wallyqs (approved):
lgtm

PR #1116, opened 2020-10-26
[FIXED] Command line `-file_slice_max_bytes 0` was not working

kozlovic:
It was impossible to set the file slice max bytes to 0. The default
64MB value would be used. Note that setting it in the configuration
file would work, only the command line would not.


derekcollison (approved):
LGTM

PR #1128, opened 2020-11-19
[CHANGED] User cluster ID as part of internal subscriptions

kozlovic:
We were using a NUID for subjects such as `_STAN.pub.<nuid>.>`
when running in standalone mode without channel partitioning.

In all other cases, we used the cluster ID in place of the nuid.
This change will make use of the cluster ID in all cases. This
will apply only to new deployment, meaning that existing stores
with NUID in the subjects will still be used.


derekcollison (approved):
LGTM

PR #1136, opened 2021-01-05
[ADDED] ReplaceDurable option

### LINKED ISSUE #1135: "duplicate durable registration" error after "subscribe request timeout"
I've been testing how my stan clients, written using nats.c, behave in the face of network problems introducing high packet loss (which, unfortunately, happens on rare occasions between two of our geographically dispersed sites). During my testing I've found that on some occasions, when resubscribing after a connection failure, the subscription request will fail with "subscribe request timeout" then, after retrying, will fail with "duplicate durable registration". All subsequent resubscribe attempts will fail with that message for the duration of the connection.

The reconnect logic my stan clients use is pretty straight forward and believe matches the recommendation:
* Set the ConnectionLostHandler on the stan connection
* The ConnectionLostHandler will:
  * Destroy all subscriptions
  * Destroy the connection
  * Attempt to reestablish the connection until it succeeds
  * Attempt to resubscribe each subscription until it succeeds unless the subscription fails due to NATS_CONNECTION_CLOSED in which case we stop everything and let the ConnectionLostHandler run again.

I'm simulating an unreliable network by using Linux's [NetEm](https://man7.org/linux/man-pages/man8/tc-netem.8.html). I've configured the network interface to rate limit to 150kbit and 40% packet loss (unrealistically high but appears to be quite good at hitting many possible failure cases when running for hours/days at a time). Most of the time the re-connection logic works correctly, except, of course, for the originally stated problem.

I haven't confirmed this but it appears as if the SubscriptionRequest message gets to the server but the SubscriptionResponse message doesn't get back to the client before the ConnectionWait period elapses causing a situation where the server thinks the client has a valid subscription but the client disagrees. This traps the client in a state where the subscription request can never be fulfilled until the connection stops and starts again.

Out of curiosity I tested the same scenario with nats-replicator. It too will get trapped in a "duplicate durable registration" error loop until the connection dies again.

Setting the ConnectionWait time to something higher (I've been using 10 seconds) significantly reduces the occurrence of the problem but does not entirely eliminate it.

The only work around I've been able to find is to simply destroy the connection entirely and reconnect when this error is encountered. However, that seems like a fairly drastic solution. Is there a better workaround for this?

kozlovic:
When the subscription request for a durable subscription times out
or fail on the client side, but it was accepted in the server, then
if the application tries to restart the subscription request again
it will fail with a "duplicate durable subscription" error until
the connection is closed.

This new option allows the user to decide how the server should behave
when processing a duplicate durable subscription. If disabled, the
default, it behaves as described above, that is, it will reject
the second subscription request and return the "duplicate durable"
error.
If enabled, if the server detects that this is a duplicate, it will
close the active one and accept the new one. It is a suspend followed
by a resume.

From the client perspective, if this is done in the context of #1135,
then everything works well since the original subscription in the
client was actually not started due to subscription request failure.
However, if user try to create multiple duplicate durable subscriptions
for subscription requests (Subscribe() calls) that did not fail, then
their application will not be notified that the subscriptions that are
being replaced are replaced, but they will simply stop receiving messages
on those.

Resolves #1135

### DIFF
--- a/nats-streaming-server.go
+++ b/nats-streaming-server.go
@@ -52,6 +52,7 @@ Streaming Server Options:
           --encrypt <bool>               Specify if server should use encryption at rest
           --encryption_cipher <string>   Cipher to use for encryption. Currently support AES and CHAHA (ChaChaPoly). Defaults to AES
           --encryption_key <string>      Encryption Key. It is recommended to specify it through the NATS_STREAMING_ENCRYPTION_KEY environment variable instead
+          --replace_durable <bool>       Replace the existing durable subscription instead of reporting a duplicate durable error
 
 Streaming Server Clustering Options:
     --clustered <bool>                     Run the server in a clustered configuration (default: false)
--- a/server/clustering_test.go
+++ b/server/clustering_test.go
@@ -8031,3 +8031,53 @@ func TestClusteringAddRemoveClusterNodesWithBootstrap(t *testing.T) {
 		}
 	}
 }
+
+func TestClusteringDurableReplaced(t *testing.T) {
+	cleanupDatastore(t)
+	defer cleanupDatastore(t)
+	cleanupRaftLog(t)
+	defer cleanupRaftLog(t)
+
+	// For this test, use a central NATS server.
+	ns := natsdTest.RunDefaultServer()
+	defer ns.Shutdown()
+
+	// Configure first server
+	s1sOpts := getTestDefaultOptsForClustering("a", true)
+	s1sOpts.ReplaceDurable = true
+	s1 := runServerWithOpts(t, s1sOpts, nil)
+	defer s1.Shutdown()
+
+	// Wait for it to bootstrap
+	getLeader(t, 10*time.Second, s1)
+
+	// Configure second server.
+	s2sOpts := getTestDefaultOptsForClustering("b", false)
+	s2sOpts.ReplaceDurable = true
+	s2 := runServerWithOpts(t, s2sOpts, nil)
+	defer s2.Shutdown()
+
+	// Configure third server.
+	s3sOpts := getTestDefaultOptsForClustering("c", false)
+	s3sOpts.ReplaceDurable = true
+	s3 := runServerWithOpts(t, s3sOpts, nil)
+	defer s3.Shutdown()
+
+	testDurableReplaced(t, s1)
+
+	s1.Shutdown()
+	newLeader := getLeader(t, 2*time.Second, s2, s3)
+	c, err := newLeader.lookupOrCreateChannel("foo")
+	if err != nil {
+		t.Fatalf("Error looking up channel: %v", err)
+	}
+	if subs := c.ss.getAllSubs(); len(subs) != 0 {
+		t.Fatalf("Expected 0 sub, got %v", len(subs))
+	}
+	c.ss.RLock()
+	lenDur := len(c.ss.durables)
+	c.ss.RUnlock()
+	if lenDur != 1 {
+		t.Fatalf("Expected 1 durable, got %v", lenDur)
+	}
+}
--- a/server/conf.go
+++ b/server/conf.go
@@ -199,6 +199,11 @@ func ProcessConfigFile(configFile string, opts *Options) error {
 				return err
 			}
 			opts.NKeySeedFile = v.(string)
+		case "replace_durable", "replace_durables", "replace_duplicate_durable", "replace_duplicate_durables":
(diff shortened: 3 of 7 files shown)


derekcollison (approved):
LGTM

kozlovic:
@adklager @yadvlz @etrochim: I wonder if you could have a go with this PR and see if that helps the issue you have observed. If not, that's ok, I will merge and this will be available in the next release.

adklager:
I did a `go get github.com/nats-io/nats-streaming-server@be195d378bb36e4cb42830331e2d7da2e56a931a`
then created a `stan.conf` file with the contents:
```
replace_durable: true
```
and instantiated the server with 

`nats-streaming-server --sc /home/aklager/stan.conf ...`

With this build and setting, I am no longer able to reproduce the initial request timeout error that led to the subsequent duplicate error. This is not the outcome I expected, though it does seem to be working now in a scenario that it previously did not.

kozlovic:
Thank you for testing it. Some comments:

> With this build and setting, I am no longer able to reproduce the initial request timeout error that led to the subsequent duplicate error. 

To be clear, this PR does not solve any timeout that your application may have experienced when creating the subscription. Some users had the same "timeout" issue and consequent "duplicate" error because they were on purpose making their network being very lossy, etc.. (see https://github.com/nats-io/nats-streaming-server/issues/1135)

> This is not the outcome I expected, though it does seem to be working now in a scenario that it previously did not.

Again, this PR simply allows users to call Subscribe() from the same connection for the same durable without getting the "duplicate durable" error and instead replace the previous with the latest. If this option is enabled, users should make sure that they are not "double" subscribing unless the previous subscription request had failed. The application would NOT receive the message twice though.

adklager:
I am basically putting a breakpoint here https://github.com/nats-io/go-nats/blob/73ffc26dfe70cc3b217df38d209c83b6d07eaf3c/nats.go#L2623 to cause a user timeout failure on the initial request, and then on a subsequent call would see the duplicate error. With the new build, I no longer see the initial timeout, and therefore never make a second call. Any reason why that initial call doesn't still timeout as expected? Sounds like this is not a valid repro.

adklager:
> To be clear, this PR does not solve any timeout that your application may have experienced when creating the subscription. 

But it did. That's what's confusing :) ..and I depended on that error to "enable" the duplicate error.

kozlovic:
Your breakpoint is too late. The request has already been sent, and the message is likely to be available when you resume execution. So the select will get the message and return it.

How did you get the timeout error in the first place, which lead you to open the issue since the second subscribe would have caused a duplicate? Was it just an error that happened one day - maybe the server was overloaded and did not send reply on time, or there was a network disconnect at the bad time, etc..

adklager:
The "real" error happened on a heavy loaded server that took longer than the timeout to reply. I've since been able to reproduce though using the breakpoint- which sounds like it was a race condition, and I got lucky. Do you have a suggestion as how to generate the initial error?

kozlovic:
If your server is local and you can "play" with it, you could put it in paused state (ctrl+z), and have the client send the request. That request will timeout. You can then resume the server "fg" (for foreground).

Or if you can modify the code/debugger runtime, set `err = nats.ErrTimeout` after this line: https://github.com/nats-io/stan.go/blob/5e2953cdd5b1dc02070fec826b0752ea18afbe0d/sub.go#L286

But to be fair, I was more asking for you guys to make sure that this option is a viable solution and does not break any of the existing deployment that you have.

adklager:
Thanks for the suggestions.

> But to be fair, I was more asking for you guys to make sure that this option is a viable solution and does not break any of the existing deployment that you have.

Got it - works for me :+1: 

As a trivial future request, we exclusively use command line arguments and not a config file, so that parity would be useful.

adklager:
> Or if you can modify the code/debugger runtime, set `err = nats.ErrTimeout` after this line: https://github.com/nats-io/stan.go/blob/5e2953cdd5b1dc02070fec826b0752ea18afbe0d/sub.go#L286

Doing this on the first request does still cause a duplicate durable registration error on subsequent requests.

kozlovic:
> Doing this on the first request does still cause a duplicate durable registration error on subsequent requests.

If you set the `err` to timeout only on the first attempt, and then your app tries to subscribe again, it should work. Of course, only if the server has the new option enabled. I just did it and it works for me.

PS: I am adding the `-replace_durable` as a command line option as you suggested, so will update the PR soo.

derekcollison (approved):
LGTM

PR #1139, opened 2021-01-08
[FIXED] Possible race causing ack'ed message to be redelivered

### LINKED ISSUE #1138: Server may redeliver a message that has just been acknoweldged
Normally, applications should acknowledge all messages that they receive, even if it is a redelivered message.
But if an application keeps track of which message the application has acknowledged and never call Ack() again on the same message, and if a race happens in the server, a message may be redelivered and added to the pending list and will always be redelivered.
Again, this would be a problem only for applications that use manual ack mode and do not ack a redelivered message that it believes has already been acknowledged.

kozlovic:
During redelivery, a go routine gathers all pending messages under
a lock. However, the actual send of messages is done one by one
and by then, one of the message in the list of messages to redeliver
may have been acknowledged. So during the send, under the sub's lock,
we need to check that if a message marked as redelivered is no longer
in the pending list, it should not be sent to the client.

Also fixed an unrelated data race that was just found running tests.

Resolves #1138

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -3550,6 +3550,13 @@ func (sub *subState) makeSortedPendingMsgs() []*pendingMsg {
 	return results
 }
 
+// Returns true if the message is in the pending map, false otherwise.
+// Lock is held on entry.
+func (sub *subState) isMsgStillPending(m *pb.MsgProto) bool {
+	_, pending := sub.acksPending[m.Sequence]
+	return pending
+}
+
 func qsLock(qs *queueState) {
 	if qs != nil {
 		qs.Lock()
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -3721,23 +3730,28 @@ func (s *StanServer) performAckExpirationRedelivery(sub *subState, isStartup boo
 		// otherwise this could cause a message to be redelivered to multiple members.
 		if !isClustered && qs != nil && !isStartup {
 			qs.Lock()
+			sub.Lock()
+			skip := !sub.isMsgStillPending(m)
```
ColinSullivan1, line 3734: very minor nitpick, take it or leave it, but could use `msgPending := sub.isMsgStillPending(m)` and avoid the `!`s.  It might read easier.

kozlovic: Will do, and actually should not try to send if not pending.. so thanks!

ColinSullivan1 (approved):
LGTM.

ColinSullivan1 (approved):
LGTM.

PR #1140, opened 2021-01-08
[IMPROVED] Bulk insert of messages

### LINKED ISSUE #1132: sql store performance enhancement
hi, i read the nats-streaming-server sql store souce code.  it seems that, a sql transaction is used (begin, preparedStmts, commit) to bulk insert messages.
but , i think this kind of bulk insertion [insert into tb values(1, 'a'),(2,'b'),(3,'c')] is more performant than the above.

kozlovic:
A new configuration `bulk_insert_limit` switches the server from
the current insertion of messages within a SQL transaction, to
an "INSERT INTO MESSAGES () VALUES (),(),..." which can speed up
performance by several folds.

The server still may perform regular insert within transaction
if the limit is deemed too low.

This new configuration parameter is not enabled by default. It
needs to be explicitly set, either in configuration file or
from command line `--sql_bulk_insert_limit <number here>`.

Resolves #1132

### DIFF
--- a/stores/sqlstore.go
+++ b/stores/sqlstore.go
@@ -185,6 +185,10 @@ const (
 	// Limit of number of messages in the cache before message store
 	// is automatically flushed on a Store() call.
 	sqlDefaultMsgCacheLimit = 1024
+
+	// If bulk insert limit is set, the server will still insert messages
+	// using tx if the limit is below this threshold.
+	sqlMinBulkInsertLimit = 5
 )
 
 // These are initialized based on the constants that have reasonable values.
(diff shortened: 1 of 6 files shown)


stores/sqlstore.go
```diff
@@ -1664,8 +1699,80 @@ func (ms *SQLMsgStore) flush() error {
...
-		ms.createExpireTimer()
+	return nil
+}
+
+// Insert messages with INSERT INTO MESSAGES () VALUES (),(),()...
+// This is faster than the original insert with transactions.
+// It is done only if user configures the BulkInsertLimit option.
+// Lock held on entry.
+func (ms *SQLMsgStore) bulkInsert(limit int) error {
+	const insertStmt = "INSERT INTO Messages (id, seq, timestamp, size, data) VALUES "
```
aricart, line 1710: why two different strings for the template?

kozlovic: Sorry, I am not sure I understand. There is only 1 "insert" statement value, which is this one. The values however are different because mysql wants: "(?,?,?,?,?)", while postgres wants: "($1, $2, $3, $4, $5)".

aricart (approved):
LGTM - sql parser for the win.

PR #1153, opened 2021-02-13
[FIXED] Avoid possibly dropping client protocol requests

kozlovic:
The NATS Streaming server uses NATS subscriptions to receive protocols
from the streaming clients. Except for the publish requests, these
internal subscriptions did not set the limits to "unlimited", which
could possibly lead to situation where some requests would be dropped
while other would not (if the server was slowing down to a point where
the NATS library would start to drop messages if the those internal
subscriptions queues were at the limit).

Of course, it is always possible that protocols be missed because
clients may be connected to a different server, and there could be
network issues, etc.. but this will reduce the risk.


derekcollison (approved):
LGTM

PR #1154, opened 2021-02-13
[ADDED] Subscriptions count in some monitoring endpoints

kozlovic:
The detail of a client and channel will now have a `subs_count`
field that is the count of subscriptions, regardless if the
option to ask for subscriptions details is requested.

Even when the subscriptions details is requested, the count field
allows to quickly see how many subscriptions are present in the
array.


derekcollison (approved):
LGTM

PR #1155, opened 2021-02-13
[IMPROVED] Clustering: performance for fragmented RAFT logs

kozlovic:
An option for the `go.etcd.io/bbolt` library allows to set the type
of freelist merge functions, from the default `FreelistArrayType` to
an optional `FreelistMapType`. The doc explains:

```
Array which is simple but endures dramatic performance degradation
if database is large and framentation in freelist is common.
The alternative one is using hashmap, it is faster in almost all
circumstances.
```

However, although most of the options can be set through options
or by setting to the bolt.DB object itself, the `DB.FreelistType`
field was set after opening the bolt.DB object, which was too
late and would therefore always default to the "array" merge
function.
This PR now correctly creates bolt.Options and pass those to
the bolt.Open() call so that the options are properly used.

### DIFF
--- a/server/raft_log.go
+++ b/server/raft_log.go
@@ -64,13 +64,15 @@ func newRaftLog(log logger.Logger, fileName string, sync bool, _ int, encrypt bo
 		fileName: fileName,
 		codec:    &codec.MsgpackHandle{},
 	}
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

PR #1156, opened 2021-02-17
[ADDED] Configuration parameters for RAFT's BoltDB store

kozlovic:
Two new options can be configured:

`bolt_free_list_sync: <bool>` will cause free list to be sync'ed
on write. The default is `false` which means no sync.
You would enable this option if startup time matters more than
write performance because with this option, the DB does not need
to perform a full database re-sync.

`bolt_free_list_array: <bool>` switches to the "array" type of
free list, which may suffer for large databases with fragmented
free list. Up to v0.20.0 (included), this was the default but
PR #1155 made the "hashmap" free list type now the default.
This option allows you to revert to previous behavior.


derekcollison (approved):
LGTM

PR #1157, opened 2021-02-17
[FIXED] Reduce contention with client heartbeats and prevent panic

kozlovic:
A new connection is new to handle server-to-client sending of heatbeats.
This could cause contention with the "general" connection that was
previously used since this connection is used for handling other
protocols.
In situations where there are lots of clients and the heartbeat settings
are low, this could cause contention.

Also fixed a bug that could have caused a panic should the server handle
a PING from the client while that client had just been closed.


derekcollison (approved):
LGTM

PR #1158, opened 2021-02-18
[IMPROVED] Reduce use of locking in raftlog.go

kozlovic:
raftLog is a wrapper to bolt's DB database. I think that we
can remove some of the raftLog's locking since bolt's DB itself
will ensure locking.

I am not expecting much of a performance boost with this PR, but
it has been observed contention between GetLog and StoreLogs in
some cases, so that may help.


derekcollison (approved):
LGTM

PR #1159, opened 2021-02-18
[ADDED] Configuration parameters for RAFT's BoltDB store

kozlovic:
This replaces PRs #1155 and #1156.

In 1155, I made boltdb's default freelist type to be map instead
of array, and in 1156 added options to enable array type.

This PR keeps the default to array and allow user to change to
map instead (so the config param is now `bolt_free_list_map` instead
of `bolt_free_list_array` that was introduced in 1156.

The reason is to minimize possible unexpected behavior in deployments
that did not have any issue.
I am not sure of all the extent (performance wise) of changing to
map instead of array and would rather it to be an explicit choice
made by users, and only if they observe a performance degradation
that can be attributed to the freelist.

We can always make it the default later.

### DIFF
--- a/server/clustering.go
+++ b/server/clustering.go
@@ -111,14 +111,13 @@ type ClusteringOptions struct {
 	// speed up recovery since there is no need for a full database re-sync.
 	BoltFreeListSync bool
 
-	// BoltFreeListArray sets the backend freelist type to "array".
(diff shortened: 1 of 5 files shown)


derekcollison (approved):
LGTM

PR #1160, opened 2021-02-18
[ADDED] Clustering: ability to create NATS connections to nodes

kozlovic:
Normally, the server has a single NATS connection and subscribes
to a subject where other nodes can submit requests to "connect" to it.
When a remote connects, a new subscription on an inbox is created on
both sides and they use their single "raft" NATS connection to communicate.
If node "A" connects to both "B" and "C" it will have two subscriptions
and two "outbox" subjects (on per remote node) to which send data to.

With the new option `nodes_connections` enabled, NATS connection(s)
will be created per remote node. This should help with performance
and reduce contention.
The RAFT transport is pooling connections, so there may be more than
one connection per remote node.


derekcollison (approved):
LGTM

PR #1161, opened 2021-02-22
[FIXED] Fail faster subscription requests for invalid clients

kozlovic:
When the server detects that a client is not responding to heartbeats,
it closes it, however, this was done with a barrier that ensures
that all pending protocols on all internal subscriptions are first
processed.
In the case of this close (as opposed to processing a client close
request), this is not welcome because it may causes the server
to process lots of subscription requests that were sent by that
client - maybe in response to timeouts for a server already having
issues handling the load.

The server will now forcefully close the client (without the use
of the barrier) and the detection of an invalid client is done
sooner in the handling of the subscription request.

Also, the internal subscription for client subscription requests
has been reverted to be limited (with possibly dropping requests).

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2778,7 +2778,10 @@ func (s *StanServer) initInternalSubs(createPub bool) error {
 		}
 	}
 	// Receive subscription requests from clients.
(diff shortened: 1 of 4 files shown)


derekcollison (approved):
LGTM

PR #1162, opened 2021-02-25
[FIXED] Support closing of subscriptions with inbox instead of ack inbox

kozlovic:
This will allow the client library to issue a sub close request
on a subscribe timeout. Since the library won't have the AckInbox,
it will send a sub close request with Inbox and if the server
fails the lookup with ack inbox, falls back to lookup by inbox.


derekcollison (approved):
LGTM

PR #1167, opened 2021-03-02
[FIXED] Clustering: possible panic with RAFT transport

kozlovic:
If there was a partial and what was in the pending buffer is less
than the size of the caller's buffer in Read(), we would get a
panic.


derekcollison (approved):
LGTM

PR #1169, opened 2021-03-05
[FIXED] Clustering: possible slow consumer errors

kozlovic:
This was introduced in v0.21.0 when I moved to sync subscriptions
and missed the fact that the pending limits do not apply to those
type of subscriptions.
Switching to async sub and adding buffers to a pending list.
These are consumed in the Read() call made by the RAFT library.

### DIFF
--- a/server/raft_transport.go
+++ b/server/raft_transport.go
@@ -72,64 +72,131 @@ type natsConn struct {
 	remoteAddr natsAddr
 	sub        *nats.Subscription
 	subTimeout time.Duration
-	pending    []byte
+	pendingH   *pendingBuf   // head of pending buffers list
+	pendingT   *pendingBuf   // tail of pending buffers list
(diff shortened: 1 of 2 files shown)


server/raft_transport.go
```diff
@@ -72,64 +72,131 @@ type natsConn struct {
...
 	closed     bool
 	parent     *natsStreamLayer
 }
 
+type pendingBuf struct {
+	buf  []byte
+	next *pendingBuf
+}
+
+var pendingBufPool = &sync.Pool{
```
derekcollison, line 88: I see you use these, they help?

derekcollison (approved):
LGTM

server/raft_transport.go
kozlovic: Honestly, with tests I have done, not really. But I have been using some in SQL code, etc..

PR #1171, opened 2021-03-05
[FIXED] Added missing cluster node ID in serverz endpoint

### LINKED ISSUE #1166: Reducing number of nodes in nats cluster
Dear Nats Community,

We have 7 nodes in Nats cluster. Everything is allright, no issues at all. But we do see that the nodes are not loaded at all and we can serve the current number of customers with 3 nodes.
Is it possible anyhow to decrease the number of Nats nodes by 4? If we simply shutdown the four nodes, then the cluster will end up with the lack of quorum.

kozlovic:
Resolves #1166


ColinSullivan1 (approved):
LGTM.

PR #1177, opened 2021-04-14
[FIXED] Clustering: memory growth

### LINKED ISSUE #1176: Possible memory leak in v0.21.1
Hi,

First of all let me thank you, we've been using nats streaming for almost 3 years and been quite satisfied with it. Sorry to see the deprecation notice.

We set up NATS streaming in clustering mode with file store. After upgrading our dev system from 0.20.0 to 0.21.1 we seen  strange memory growth, which looks like leaking memory. We have running 3 replicas on k8s and with the earlier versions memory usage was quite stable for us (in the current production cluster it's almost a straight horizontal line around ~100 Mb for 30 days back). After upgrading to version 0.21.1 memory usage constantly increasing  until we let it, without memory limits it went up to several GB or with the limit it gets OOM killed regularly.

if i run the basic clustering test with increasing the number of messages to a big number and adding one millisec delay between them: https://github.com/nats-io/nats-streaming-server/blob/master/server/clustering_test.go#L709
triggers 4x larger memory consumption on tag v0.21.1 than on v0.20.0

kozlovic:
Changes in v0.21.0 made heavy use of time.After() in a select that
caused an unacceptable memory growth.
Replaced with creating a timer once and using stop/reset kept
the memory usage in check.

Resolves #1176

### DIFF
--- a/server/raft_transport.go
+++ b/server/raft_transport.go
@@ -75,6 +75,7 @@ type natsConn struct {
 	pendingH   *pendingBuf   // head of pending buffers list
 	pendingT   *pendingBuf   // tail of pending buffers list
 	ch         chan struct{} // to send notification that a buffer is available
+	tmr        *time.Timer
 	outbox     string
 	closed     bool
 	parent     *natsStreamLayer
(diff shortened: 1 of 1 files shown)


kozlovic:
Seems like @hullarb that reported the issue has been running with server built from this branch and has seen that memory is stable with this fix. See comment here: https://github.com/nats-io/nats-streaming-server/issues/1176#issuecomment-819440709

derekcollison (approved):
LGTM

PR #1190, opened 2021-05-17
[FIXED] Subscription redelivery count map was not cleaned-up

kozlovic:
PR #997 introduced a map to keep track of how many times a message
was redelivered to a subscription (or a queue group). However, sequences
added to this map was never removed when the ACK was processed from
the subscription!


derekcollison (approved):
LGTM

PR #1191, opened 2021-05-18
[ADDED] Make http handlers public

BlizzTom:
This matches how the nats-server exposes the http monitoring endpoints as public

For people using this as a go package, allows for attaching to an existing http server instead of having to expose a duplicate http server endpoint

Discussed with @kozlovic over slack


kozlovic (approved):
LGTM. Thank you for the contribution!

PR #1201, opened 2021-07-12
[FIXED] Possible redelivery from node that lost leadership

kozlovic:
I have added a test with a sleep in performRedeliveryOnStartup
that would demonstrate the issue when this function was started
as a go routine. The sleep/test is a bit useless with new code
since this is invoked in place, but will catch the issue if
in the future this function is again invoked as a go routine.


aricart (approved):
LGTM

PR #1203, opened 2021-07-20
[FIXED] Clustering: snapshot queue group's last_sent

kozlovic:
In some situations where a member has fallen behind and other
members left the group, a snapshot would have the lone member
with a lower last_sent, which could cause older messages to
be redelivered.


derekcollison (approved):
LGTM

PR #1204, opened 2021-07-20
[FIXED] SQL: sub last_sent could end-up higher than store last_seq

kozlovic:
With message expiration, after some sequence of events, it could
be that a subscription's last_sent sequence would be higher
than the store's last_seq.


derekcollison (approved):
LGTM

PR #1205, opened 2021-07-21
Ensure RAFT connection are closed if transport is closed

kozlovic:
Some races could leave dialed/accepted connections to be added
to the list (and not closed) on some tests shutdown, which could
cause some other tests to fail because of unexpected reconnects.


wallyqs (approved):
LGTM

PR #1213, opened 2021-09-28
[FIXED] last_sent may be 0 in monitor endpoint on followers

### LINKED ISSUE #1189: Monitoring desync on Leader change - Subscription last_sent = 0
Setup:
- nats-streaming-server 0.21.1
- 5 Nodes running in Containers as AWS ECS Tasks
- we have recently expanded Cluster in accordance with - https://github.com/nats-io/nats-streaming-server/issues/1181

When the Cluster Leader is changed the Monitoring endpoints on the Peers show incorrect Subscription state (last_sent = 0). That touches the Channels where we had an Error "leadership lost while committing log". Later the Monitoring will be resyncronized, if a new message is published to the channel. If there are no new incoming messages - no sync happens and last_sent stays at 0. Still the monitoring endpoint of the Leader has a correct state.

### Expected Behaviour
After Leader Change monitoring endpoints of non-leader Nodes reflect the correct Subscription State.

### Actual Behaviour
After Leader Change monitoring endpoints of non-leader Nodes report last_sent = 0 by the Subscriptions, where an Error "leadership lost while committing log" occurred

kozlovic:
If a queue subscription is replacing a timed out one, the
leader would update its runtime version of the subscription's last_sent,
however, this would not be visible on the followers' monitoring
endpoint(s) until new messages were delivered.

Resolves #1189

### DIFF
--- a/server/clustering_test.go
+++ b/server/clustering_test.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"net"
+	"net/http"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #1220, opened 2021-11-10
[FIXED] Clustering: possible wrong pending count

kozlovic:
This is a regression introduced by PR #1213 which is in v0.23.0
(the latest release).

When a queue subscription left the group and its last_sent was the
same than the group's last_sent, the remaining queue member should
have its last_sent updated. To do that, the replication of a "sent"
event for this sequence was issued, which was wrong since it would
possibly add a pending on the remaining queue sub. Moreover, this
did not account for last_sent being 0, which with the aforementioned
behavior, a "sent" of sequence 0 would be replicated causing the
remaining queue sub to show a pending_count of 1.

The fix for both wrong pending_count and original last_sent issue
is to have all nodes (leader and followers) detect that when a
member is removed, if that member's last_sent value was the one
of the queue group, then update the runtime version of the remaining
queue member.

### DIFF
--- a/server/clustering_test.go
+++ b/server/clustering_test.go
@@ -8384,123 +8384,140 @@ func TestClusteringSnapshotQSubLastSent(t *testing.T) {
 	}
 }
 
-func TestClusteringMonitorQueueLastSentAfterHBTimeout(t *testing.T) {
-	resetPreviousHTTPConnections()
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #1224, opened 2021-11-17
[FIXED] Clustering: possible wrong pending_count and is_stalled on followers

kozlovic:
If a queue member receives a message and does not ack it within
the AckWait interval, this message may be redelivered to a different
member. If the original member then acknowledge the message, the
server finds it in the other member and remove it from the pending
list. This would cause the second member to show a wrong pending_count
and possibly is_stalled values in monitoring of the followers.


derekcollison (approved):
LGTM

PR #1237, opened 2022-01-30
[FIXED] Possible panic on message redelivery

kozlovic:
If the message redelivery callback fired and a client was marked
with failed heartbeat and the subscription is closed (likely
due to the server closing connection due to failed HBs), there
was a race that could cause a panic trying to reset a timer
that has been set to nil.


derekcollison (approved):
LGTM

PR #1259, opened 2022-07-15
[ADDED] FileStore: RecordSizeLimit to limit single record read size

### LINKED ISSUE #1255: High memory consumption while recovering damaged data file (in scenario in-file storage)
High memory consumption while recovering damaged data file (in scenario in-file storage)
In case, when (for unknown reason) one of data file was damaged (for example: bad EOF mark), server make a note in log file:

{"level":"info","time":"2022-06-23T13:14:37+03:00","message":"Server is ready"}
{"level":"info","time":"2022-06-23T13:14:38+03:00","message":"STREAM: Recovering the state..."}
{"level":"error","time":"2022-06-23T13:14:43+03:00","message":"STREAM: Verification of last message for file \"C:\\\\ProgramData\\\\Some_folder\\\\NSS\\\\data\\\\Some_file.Log\\\\msgs.1.dat\" failed: unable to read last record: unexpected EOF"}
{"level":"error","time":"2022-06-23T13:14:43+03:00","message":"STREAM: Error with index file \"C:\\\\ProgramData\\\\Some_folder\\\\NSS\\\\data\\\\Some_file.Log\\\\msgs.1.idx\": Verification of last message for file \"C:\\\\ProgramData\\\\Some_folder\\\\NSS\\\\data\\\\Some_file.Log\\\\msgs.1.dat\" failed: unable to read last record: unexpected EOF. Truncating and recovering from data file"}

Memory usage exceeds more than 1,6 Gb instead of ~30 Mb in normal operating mode.

image

Problem gone by deleting the damaged file or restarting server, but in such case damaged file continues to exist and other software, whose using this file to operate, cannot start and work.
This sutuation occurs both on Windows and Linux hosts, including 0.24.6 software version.

kozlovic:
In case of memory corruption, it is possible that the record size
is way greater than it should, which would cause the server to
create a buffer of the wrong size in the attempt to read the
record. This new option will limit how big the buffer needed
to read the record from disk can be.

Resolves #1255

### DIFF
--- a/stores/filestore.go
+++ b/stores/filestore.go
@@ -196,6 +196,11 @@ type FileStoreOptions struct {
 	// by setting DoSync to false.
 	// Setting AutoSync to any value <= 0 will disable auto sync.
 	AutoSync time.Duration
+
+	// RecordSizeLimit defines the maxmimum size of a record that can be read
+	// from disk. Should a record corruption occur, this will prevent the server
+	// to allocate more memory than this value. Expressed in bytes.
+	RecordSizeLimit int
 }
 
 // This is an internal error to detect situations where we do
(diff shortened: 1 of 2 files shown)


aricart (approved):
Lgtm

PR #1267, opened 2022-08-02
[FIXED] Recover channels even if server information not recovered

### LINKED ISSUE #1263: no recovered state when missing server.dat file
noticed that if server.dat file (in data/ dir) is absent, the server doesn't attempt to recover its state  
can see messages in logs:  
`STREAM: Recovering the state...`
`STREAM: No recovered state`

but, no matter what, server.dat file is being created and on the next server relaunch the recovery process starts as expected

kozlovic:
A user reported that the server.dat was missing, which means that
the server did not recover channels (assuming that this was a
fresh start, without any state). This caused issues later on
since if an app publishes a message to an existing (non recovered)
channel, then messages would be added to the channel but with
a reset of the sequence.

The server will now warn of the situation, initialize the server
state and then recover the rest of the state (channels, etc..)

Resolves #1263

### DIFF
--- a/server/clustering_test.go
+++ b/server/clustering_test.go
@@ -6915,7 +6915,7 @@ func TestClusteringRestoreSnapshotWithDifferentVersionsOfSameChannel(t *testing.
 }
 
 func TestClusteringSQLMsgStoreFlushed(t *testing.T) {
-	if !doSQL {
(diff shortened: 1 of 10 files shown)


aricart (approved):
LGTM

PR #1274, opened 2022-10-11
[CHANGED] Reject channels with different case (Foo vs foo)

### LINKED ISSUE #1265: sub/pub on similar-named channels (on Windows)
some strange behavior encountered when tried to use some channels with similar names (like 'boo' and 'Boo')  
the experiment was as follows:  
```
 1) pub msg 1 to channel 'BearD'
 2) pub msg 2 to channel 'bEARd'
 3) pub msgs 3 to channel 'beard'
 4) pub msgs 4 to channel 'beard' again
 5) pub msg 5 to channel 'BEARD'

sub to all msgs (from the first one, non-durable) of the channel 'beard' - got msg 1 with seq 1 (expected 3 and 4)
sub to all msgs (from the first one, non-durable) of the channel 'BEARD' - got msg 1 with seq 1 (expected 5)
other variants give the msg 1 with seq 1 as well
but if we try to subscribe to 'bEaRd' for example, we'll got no msgs! ok!
```

so i proceeded:
```
the subscriber were stopped
 6) pub msg 6 to channel 'BearD'
then i subscribed to msgs of 'BearD' and still got only the msg 1 with seq 1 (didn't got msg 6)
 7) pub msg 7 to channel 'beard' - no reaction
 8) pub msg 8 to channel 'BearD' again... and got another one msg for the subscriber - msg 8 with seq 3 (!)
```

summary:
a) when use similar-named channels, you got only the first msg in this channel, regardless of which channel (used before) you subscribed to
b) with the subscription staying alive, you got the msgs exactly from that channel you subscribed to
c) each channel has its own seq numbering, despite the fact that everything is put in one folder (folder is named by the first channel name we published to - 'BearD')

kozlovic:
Due to the way channels are stored to persistent storage, the first spelling of a channel caused that channel to be persisted under that name.
If an application later publishes/subscribes to the same channel but with a different case (say first "Foo" then "foo"), the server would treat it as a new channel but would still store data in the first "Foo" channel, causing message sequences to be reset and inability to consume after a server restart.

From now on, the server will reject published messages or new subscriptions if it detects that a similar channel already exists with a different "case".

Resolves #1265

### DIFF
--- a/server/clustering.go
+++ b/server/clustering.go
@@ -738,6 +738,7 @@ func (r *raftFSM) lookupOrCreateChannel(name string, id uint64) (*channel, error
 			return nil, err
 		}
 		delete(cs.channels, name)
+		delete(s.channels.channelsLC, strings.ToLower(name))
 	}
 	// Channel does exist or has been deleted. Create now with given ID.
 	return cs.createChannelLocked(s, name, id)
--- a/server/server.go
+++ b/server/server.go
@@ -274,16 +274,18 @@ func (state State) String() string {
 
 type channelStore struct {
 	sync.RWMutex
-	channels map[string]*channel
-	store    stores.Store
-	stan     *StanServer
+	channels   map[string]*channel
+	channelsLC map[string]*channel
+	store      stores.Store
+	stan       *StanServer
 }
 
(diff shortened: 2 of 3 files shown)


derekcollison (approved):
LGTM

PR #1303, opened 2024-04-01
[FIXED] Documentation link [ci skip]

### LINKED ISSUE #1302: Official documentation link is broken (README)
On the README file there's a reference to the official documentation (legacy) which shows a `page not found` error now. If it's moved somewhere we should change the README file. If it's completely deleted we should delete this part from README; which I hope is not the case. It really makes sense to keep the documentation for this product available even when it is deprecated.

kozlovic:
The link to legacy link was updated.

Resolves #1302

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The NATS Streaming Server is being deprecated.  Critical bug fixes and security 
 
 ## Documentation
 
-* [Official documentation](https://docs.nats.io/legacy/stan)
+* [Official documentation](https://nats-io.gitbook.io/legacy-nats-docs/nats-streaming-server-aka-stan)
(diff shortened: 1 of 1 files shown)


kozlovic:
@derekcollison @Jarema @gcolliso A user reported the documentation link was broken and this PR fixes it, but it would conflict with PR (https://github.com/nats-io/nats-streaming-server/pull/1299) that is removing almost all links just to say that it is no longer supported and points to JetStream. So not sure if this PR is needed or not. The user states that he understands that the product is EOL but still thinks that having access to the doc is a good thing.

derekcollison (approved):
LGTM

PR #1311, opened 2025-12-31
Update go to 1.24.11

Shubham-astro:
## Description
Update Go version to 1.24.11 to resolve CVE: CVE-2025-61729


kozlovic:
We no longer update this repository since it was EOL in June 2023. See note [here](https://github.com/nats-io/nats-streaming-server?tab=readme-ov-file#warning--deprecation-notice-warning).

PS: Even if the product was still active, we would not take a PR that updates the executables, a new release would produce new assets.

